PR #14, opened 2015-07-07
Added withNameInsensitive and tests

shawndowns:
I would like to add withNameInsensitive, like withName but case insensitive. Tests added, matching withNameInsensitiveOption, and passed


lloydmeta:
Thanks ! I just pushed 1.2.3-SNAPSHOT and 1.2.3 to OSS Sonatype. It should be available soon from Maven Central.

shawndowns:
No problem :). Aside from ScalaJS, because I have no experience with that, let me know if there's anything you want help with.

PR #21, opened 2015-12-01
Added enum collections (set, map)

mauhiz:
These collections could become efficient if `Enum.indexOf` is ever optimized! For now, their only advantage is being very compact and very nerdy.


lloydmeta:
Interesting. Out of curiosity, what inspired you to make these ?

mauhiz:
Lack of day to day Scala, and a flash of inspiration during a shower :)

lloydmeta:
I think we should put these classes in a separate subproject called enumeratum-extras (currently no demand and they are not strictly necessary) and have the mutable and immutable versions in different packages (following standard lib convention)

mauhiz:
I don't think this is useful for anyone after all. :traffic_light:

PR #22, opened 2015-12-01
- Add valuesToIndex map for optimised index getting

lloydmeta:
- Add test for -1 case when no such value exists


lloydmeta:
Let's #21 achieve its advertised constant time accesses.

PR #25, opened 2015-12-29
Improved exception messages

luksow:
Enum's mangled name used in previous version is not very useful, especially when used in conjunction with frameworks such as spray or akka-http where bubbled up exceptions are shown to end users as errors. Providing enum members seems as a better idea. Let me know what you think.


lloydmeta:
@luksow this is a great idea. Thanks !

lloydmeta:
Just released 1.3.6 with this feature.

luksow:
That was quick, thanks! Keep up the good work!

PR #32, opened 2016-05-02
ReactiveMongo

lambdista:
Add support for [reactivemongo](http://reactivemongo.org/)


enumeratum-reactivemongo-bson/src/main/scala/enumeratum/EnumHandler.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Returns a BSONHandler for a given enum [[Enum]]
+   *
+   * @param enum The enum
+   * @param insensitive bind in a case-insensitive way, defaults to false
+   */
+  def handler[A <: EnumEntry]
+    (enum: Enum[A], insensitive: Boolean = false): BSONHandler[BSONValue, A] = new BSONHandler[BSONValue, A] {
+
```
lloydmeta: It might be a good idea to have stable `val reader` (passing insensitive/sensitive) and `val writer` values inside the BSONHandler so we don't have to instantiate new ones all the time when using a`BSONHandler`.

lambdista: I can do that. What do you think about making them `lazy val` instead?

lloydmeta: In this case since the readers and writers themselves seem pretty lightweight, I'd rather not incur the cost of using lazy, but I don't feel strongly either way.

lambdista:
@lloydmeta Cool! Let me know when you publish it so I'll start using it. Thank you.

lloydmeta:
@lambdista thanks for your good work :) I've just released 1.4.2 (published to Sonatype releases, should be in Maven Central soon-ish).

lambdista:
Thank you @lloydmeta!

PR #35, opened 2016-05-06
Change error message for Play JSON to be like Play

eddsteel:
In Play JSON, a small number of constant strings are used for validation errors. These can then be [localized](https://github.com/playframework/playframework/blob/9a06b9a0a6ea7d1299e61ea3c1f37a08b714297e/framework/src/play/src/main/resources/messages.default). It also enables non-Play code to use this message, combined with the path specifying the bad field's location, to generate useful messages.

This change modifies the Play JSON `Reads` for enums to use strings already [in use](https://github.com/playframework/playframework/blob/bf29d718c426d1d30d10553ad2e6b609bc1091fe/framework/src/play-json/src/main/scala/play/api/libs/json/Reads.scala#L752) in Play. Although on their own they are less informative than what's there now, they fit in better with Play and when combined with paths can be used to make more informative messages. (This is our use case. We're not using the play framework but expecting Play JSON behaviour. With current enumeratum behavior we have to prefix-match the error string to generate our errors, but match constant strings for other fields).

If you prefer I can make this an optional thing, like case-insensitivity, but I thought I'd propose this simpler change first.

### DIFF
--- a/enumeratum-play-json/src/main/scala/enumeratum/EnumFormats.scala
+++ b/enumeratum-play-json/src/main/scala/enumeratum/EnumFormats.scala
@@ -19,10 +19,10 @@ object EnumFormats {
         val maybeBound = if (insensitive) enum.withNameInsensitiveOption(s) else enum.withNameOption(s)
         maybeBound match {
           case Some(obj) => JsSuccess(obj)
-          case None => JsError(s"Enumeration expected of type: '$enum', but it does not appear to contain the value: '$s'")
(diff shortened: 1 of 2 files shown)


lloydmeta:
Hey there, I think think this change is beyond reasonable. I'm not too worried about breaking backwards compatibility because:

1. This is about changing the error case, so success cases will not change. 
2. As you mentioned, it fits how Play-JSON handles errors better.

Thanks !

PR #36, opened 2016-05-17
Added description how to integrate with Slick

geggo98:


README.md
```diff
@@ -607,6 +608,46 @@ val reader = implicitly[BSONReader[BSONValue, BsonDrinks]]
...
+  def greeting = column[Greeting]("GREETING")
+
+```
+
+If you want to represent your enum in the database with numeric IDs, just provide a different mapping. This example uses the enum of type `LibraryItem` defined in the introduction:
+
+```
+  implicit lazy val libraryItemMapper = MappedColumnType.base[LibraryItem, Int](
+    item => greeting.value,
+    id => Greeting.withValue(id)
```
lloydmeta: `LibraryItem.withValue` I believe :)

README.md
```diff
@@ -607,6 +608,46 @@ val reader = implicitly[BSONReader[BSONValue, BsonDrinks]]
 assert(reader.read(BSONInteger(3)) == BsonDrinks.Cola)
 ```
 
+### Slick integration
+
+[Slick](http://slick.lightbend.com) doesn't need a separate integration. You just have to provide a `MappedColumnType` for each database column that should be represented as an enum on Scala site.
```
geggo98: Looks like your German is better than my English!

geggo98:
OK, I fixed the typos. Please let me know, when there are additional issues.

lloydmeta:
Thanks !

geggo98:
Gern geschehen! Jederzeit wieder :-)

flevour:
@geggo98 I've done this integration and it works almost fine (ie: defined the column mapper), but I have to force cast enum items to a `Rep`, ie: in this case `.filter(_.productType = ProductType.Foo)` Slick can't lift `ProductType.Foo` to `Rep[ProductType.Foo]`, so I have to cast it manually `val productType: Rep[ProductType] = ProductType.Foo` and then use that val in the query.

How's your experience in this regard?

lloydmeta:
@flevour I've seen that happen before and the cause is the compiler inferring the type of `ProductType.Foo` to be `ProductType.Foo`, the type of the singleton (this is by design, since usually you want your type inference to be as specific as possible). Unfortunately, as a result, it is unable to find your implicit mapper which will help you lift to `Rep[ProductType]`. Your way of having an intermediate `val productType` works, but I think you can also do type ascription, as in the following:

```scala
.filter(_.productType === (ProductType.Foo: ProductType))
```

Does that work for you?

BTW, just a small correction on terminology: what you're doing is not force casting, it is called type annotation. You're simply telling the compiler about the type of object you want, and the compiler will, at compile time, verify that it makes sense. Force casting is something like this: `Producttype.Foo.asInstanceOf[Rep[ProductType]]`, which could die at run time :)

geggo98:
@flevour My code looks as follows:

1. I have a model package with several enums and case classes. The case classes consist of fields with primitive types, strings or enum types.
2. I have a DAO package with classes providing typical CRUD actions. Each CRUD action typically takes a primary key and a case class or a parameter of a case class field
3. In each DAO class there is an embedded class providing the Slick table structure. This class is a Slick table for a case class in the models package (`Table[models.SomeCaseClass]`)
4. Each DAO class contains the mappings for the enums as `implicit lazy val ... = MappedColumnType[...]` like I provided it in my description

With my structure, inserting, updating, and filtering with enums work fine. But I have more explicit types in my structure. I think this is the key difference.

geggo98:
@flevour Having to cast mapped types in filters seems to be normal in Slick and not a feature of embedding case classes built with enumeratum.

If you look into the [Slick tutorial at Scalantor.io](https://www.scalanator.io/course/slick#990), you can find an example where you must add explicit casting for a purely case object based enum with explicit mapping functions: `AlbumTable.filter(_.rating >= (NotBad : Rating))`, where `Rating` is just a sealed trait with several case objects and the mapping is built like `case 1 => Good; case 2 => NotSoGood`.

flevour:
Wow, so many goodies in this thread! Thanks for being so kind to share all these tips @geggo98 @lloydmeta !

lloydmeta:
No problem at all. I've updated the docs to take into account your question :)

PR #39, opened 2016-06-15
add PlayFormFieldEnum documentation

zeapo:
This information is missing from the readme, I had to look into the source code.


lloydmeta:
You're right, the documentation can definitely be improved and, thank you very much for this PR :)

PR #40, opened 2016-06-21
README.md : fix example with sealed abstract class

dm-tran:
Using `sealed abstract class State(override def entryName: String) extends EnumEntry`, as described in README.md, raises the following error : 

```
error: 'val' expected but 'def' found.
       sealed abstract class State(override def entryName: String) extends EnumEntry
                                            ^
```

Using `sealed abstract class State(override val entryName: String) extends EnumEntry` instead compiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ to manually override the `def entryName: String` method.
 
 import enumeratum._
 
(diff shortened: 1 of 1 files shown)


lloydmeta:
@dm-tran Sorry about this oversight. Thanks for submitting a fix !

PR #42, opened 2016-06-29
Avoid boxing when calling withValue, withValueOpt on IntEnum, ShortEnum

shanielh:
Hi,

First of all, I didn't managed to run the tests successfully on the master branch, I hope that it relates to Scala.js. 

I've added a trait called `IndexedEnum`, This trait implements `withValue`, `withValueOpt` using the apply method of the indexedValues, Since the values are indexed, This trait is specialized for short and int, so calling the methods `withValue`, `withValueOpt` wouldn't cause boxing.

~~There is one thing, I'm not sure how the macro `ValueEnumMacros.findIntValueEntriesImpl[A]` works, And what happens if there is a gap between Enum Values (e.g, first value is 5).~~

I'm not sure if you want to make `ValueEnum` specialized for int and short, So there won't be boxing when people pass the `ValueEnum` as `ValueEnum` (and not as the real type)

The `ValueEnumSpec` passed and validated the changes :+1:

### DIFF
--- a/enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/IndexedEnumMap.scala
+++ b/enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/IndexedEnumMap.scala
@@ -1,0 +1,73 @@
+package enumeratum.values
+
+/**
+ * An [[ValueEnum]] that has an optimized array access that does not uses
+ * boxing when accessing the methods [[ValueEnum.withValue()]] and [[ValueEnum.withValueOpt()]]
+ */
+sealed abstract class IndexedEnumMap[@specialized(Int, Short, Long) ValueType <: AnyVal, EntryType <: ValueEnumEntry[ValueType]](values: Seq[EntryType]) {
+
+  private[this] val existingEntriesString = values.map(_.value).mkString(", ")
+  private[this] val (arrayValues, minValue) = {
+    val indices = values.map(v => toIndex(v.value))
+
+    val minValue = indices.min
+    val maxValue = indices.max
+
+    val retVal = Array.fill[Option[EntryType]](maxValue - minValue + 1)(None)
+
+    for (value <- values) {
+      retVal(toIndex(value.value) - minValue) = Some(value)
+    }
+
+    (retVal, minValue)
+  }
+
+  /**
+   * Tries to get an [[Int]] by the supplied value. The value corresponds to the .value
+   * of the case objects implementing [[Int]]
+   *
+   * Like [[Enumeration]]'s `withValue`, this method will throw if the value does not match any of the values'
+   * `.value` values.
+   */
+  final def withValue(i: ValueType): EntryType = {
+    val index = toIndex(i) - minValue
+    if (index >= arrayValues.length || index < 0) {
+      throw new NoSuchElementException(buildNotFoundMessage(i))
+    } else {
+      arrayValues(index).getOrElse(throw new NoSuchElementException(buildNotFoundMessage(i)))
+    }
+  }
+
+  /**
+   * Converts the given value to [[Int]]
+   *
+   * @param value The value to convert
+   * @return The converted value
+   */
+  @inline def toIndex(value: ValueType): Int
+
+  /**
+   * Optionally returns an [[Int]] for a given value.
+   */
+  final def withValueOpt(i: ValueType) = {
+    val index = toIndex(i) - minValue
+    if (index >= arrayValues.length || index < 0) None else arrayValues(index)
+  }
+
+  private final def buildNotFoundMessage(i: ValueType): String = {
+    s"$i is not a member of ValueEnum ($values)"
+  }
+
+}
+
+sealed class IntIndexedEnumMap[A <: ValueEnumEntry[Int]](values: Seq[A]) extends IndexedEnumMap[Int, A](values) {
+  final override def toIndex(value: Int): Int = value
+}
+
+sealed class ShortIndexedEnumMap[A <: ValueEnumEntry[Short]](values: Seq[A]) extends IndexedEnumMap[Short, A](values) {
+  final override def toIndex(value: Short): Int = value.toInt
+}
+
+sealed class LongIndexedEnumMap[A <: ValueEnumEntry[Long]](values: Seq[A]) extends IndexedEnumMap[Long, A](values) {
+  final override def toIndex(value: Long): Int = value.toInt
+}
--- a/enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/ValueEnum.scala
+++ b/enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/ValueEnum.scala
@@ -1,15 +1,15 @@
 package enumeratum.values
 
 import enumeratum.ValueEnumMacros
-
 import scala.language.experimental.macros
 
-sealed trait ValueEnum[ValueType <: AnyVal, EntryType <: ValueEnumEntry[ValueType]] {
+sealed trait ValueEnum[@specialized(Short, Int) ValueType <: AnyVal, EntryType <: ValueEnumEntry[ValueType]] {
 
   /**
    * Map of [[ValueType]] to [[EntryType]] members
    */
   final lazy val valuesToEntriesMap: Map[ValueType, EntryType] = values.map(v => v.value -> v).toMap
+  private lazy val existingEntriesString = values.map(_.value).mkString(", ")
 
   /**
    * The sequence of values for your [[Enum]]. You will typically want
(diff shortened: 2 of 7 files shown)


shanielh:
Not sure why coverage is decreased, I haven't touched any of the tests :open_mouth:

lloydmeta:
Hi there,

Thanks for your interest in this project. For my reference, was this PR made out of concern for performance? It helps for me to know your motivation :)

> First of all, I didn't managed to run the tests successfully on the master branch, I hope that it relates to Scala.js.

Really? Can you describe what errors you had? The tests run fine on my computer and on Travis.

The PR looks reasonable and I just have a few concerns:

- Looks like when values are negative, the PR as-is would not work because the values are used as array indices
- If we have enumerations where the range is quite big, we will end up with really big arrays filled with references to None
- No tests for added lines code. This is why the coverage percentage decreased.
- Might be nice to have a few benchmarks to show this complexity is worthwhile

A somewhat contrived example of something that might not work:

```scala
sealed abstract class Stuff(val value: Int) extends IntEnumEntry
object Stuff extends IntEnum[Stuff] {
  val values = findValues
  case object Beans(-1000) extends Stuff
  case object Other(Int.MaxValue) extends Stuff
}
```

shanielh:
Hi @lloydmeta .

Yeah, The main concern was performance.

For your concerns : 

* We can add it as opt-in behavior, and not "mix in" the `IndexedEnum` to `IntEnum` and `ShortEnum` by default.
* Same as above
* I haven't added that line of code, It was there in the first place :confused: 
* I can add a test using [ScalaMeter](http://scalameter.github.io/) if you want.
```

[Log for sbt test](https://github.com/lloydmeta/enumeratum/files/338799/log.txt).

No problem, I'll add commits and squash them (if and) when you want to merge :)

**Comment by lloydmeta** [comment]
Cool I hear you on the performance concern. It might be worthwhile to add some performance tests then to show the difference in performance. If we add some benchmarks in this PR, then I would prefer we use JMH because it is battle tested by professionals ([see this thread for how easily things can be implemented wrong](https://groups.google.com/forum/#!topic/mechanical-sympathy/m4opvy4xq3U)).

Even though we can make this an opt-in behaviour for the sake of performance, it might be better to try to find a way to avoid non-intuitive behaviour that throws errors at runtime and is hard to test for in the first place. Is there some other way to achieve specialisation ?

BTW, this PR adds 72 lines of code.  For instance, see the green blocks [here](https://github.com/lloydmeta/enumeratum/pull/42/files#diff-b06bfbd455a9b7b83b17066ea3664b9bR46)

**Comment by shanielh** [comment]
Can you check in JMH the number of boxing in a specific code section? (I know we're able to do that in ScalaMeter).

We can add the "lower bound" of the array as a field and fill the array from the lower bound, That way, It would work with negative values.

**Comment by lloydmeta** [comment]
> Can you check in JMH the number of boxing in a specific code section? (I know we're able to do that in ScalaMeter).

I'm not sure about checking the number of boxing in a specific code section in JMH. It's pretty interesting that you can do that in Scalatest !

> We can add the "lower bound" of the array as a field and fill the array from the lower bound, That way, It would work with negative values.

That sounds like a reasonable approach, maybe we can go with that for now? Alternately we can maybe import the code from [Debox's Map](https://github.com/non/debox/blob/master/src/main/scala/debox/Map.scala).

> By looking at the coverage report, it says that number of covered lines in Cicle decreased - https://coveralls.io/builds/6795267

This is odd considering [this part of CirceSpec](https://github.com/lloydmeta/enumeratum/blob/master/enumeratum-circe/src/test/scala/enumeratum/CirceSpec.scala#L33-L35) should invoke that line...

**Comment by shanielh** [comment]
@lloydmeta So, work items for this PR : 

- [ ] me - Add test for negative sequences and fix it
- [ ] @lloydmeta - Read about Scalatest and see if it's good enough for you

**Comment by lloydmeta** [comment]
Scalameter sounds fine to me :)

**Comment by cb372** [comment]
If you want to avoid boxing, have you considered simply replacing `Map` with `LongMap`?

**Comment by shanielh** [comment]
@cb372 - Since that won't help with the boxing / unboxing, The `Trait` map is still of `[A, B]` and both of them are not specialized. `LongMap` may be faster than normal (hash/trie) `Map` when the key is Long, but boxing/unboxing will still happen.

@lloydmeta - I've rewrote the code, the coverage should be smaller now, I'll add tests for these, but before that, The reason I changed the way of implementation is : 

1. When deriving a trait from trait, `@specialized` isn't working good : http://axel22.github.io/2013/11/03/specialization-quirks.html
2. Now the implementation is "opt in", And if you have a long enum, You can use the new `IndexedEnumMap` too.

Actually, I can take this code and use it in my case, But it would be nice to merge this PR / even add documentation about `withValue` / `withValueOpt` and detail about their boxing and how to avoid it.

Tell me what you think about it. I can add benchmark to see that the newer way is faster than the old.

File: enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/ValueEnum.scala [hunk 1]
```diff
@@ -1,15 +1,15 @@
 package enumeratum.values
```
shanielh: This file will be reverted. including the `@specialized` annotation.

enumeratum-core/compat/src/test/scala-2.11/enumeratum/values/IndexedEnumBoxingBench.scala
```diff
@@ -0,0 +1,43 @@
...
+          } else {
+            Drinks.withValue(1)
+            LibraryItem.withValue(1)
+          }
+
+          i += 1
+        }
+      }
+    }
+  }
```
shanielh: [info] ::Benchmark IndexEnum.withValueOpt::
[info] cores: 4
[info] hostname: shani-laptop
[info] name: OpenJDK 64-Bit Server VM
[info] osArch: amd64
[info] osName: Linux
[info] vendor: Oracle Corporation
[info] version: 25.91-b14
[info] Parameters(useIndexed -> false): Map(scala.runtime.BoxesRunTime.boxToLong(long) -> 0, scala.runtime.BoxesRunTime.boxToShort(short) -> 100005, scala.runtime.BoxesRunTime.boxToCharacter(char) -> 0, scala.runtime.BoxesRunTime.boxToBoolean(boolean) -> 0, scala.runtime.BoxesRunTime.boxToFloat(float) -> 0, scala.runtime.BoxesRunTime.boxToDouble(double) -> 0, scala.runtime.BoxesRunTime.boxToInteger(int) -> 100005, scala.runtime.BoxesRunTime.boxToByte(byte) -> 0)
[info] Parameters(useIndexed -> true): Map(scala.runtime.BoxesRunTime.boxToLong(long) -> 0, scala.runtime.BoxesRunTime.boxToShort(short) -> 0, scala.runtime.BoxesRunTime.boxToCharacter(char) -> 0, scala.runtime.BoxesRunTime.boxToBoolean(boolean) -> 0, scala.runtime.BoxesRunTime.boxToFloat(float) -> 0, scala.runtime.BoxesRunTime.boxToDouble(double) -> 0, scala.runtime.BoxesRunTime.boxToInteger(int) -> 0, scala.runtime.BoxesRunTime.boxToByte(byte) -> 0)

cb372:
Sorry for my very terse comment, I was on my phone. I meant something like this. (Untested, probably doesn't compile.)

```scala
import scala.collection.immutable.LongMap

// Type class for things that can be treated as a Long
trait ToLong[A] { def toLong(value: A): Long }
object ToLong {
  implicit val longToLong  = new ToLong[Long] { def toLong(value: Long) = value }
  implicit val intToLong   = new ToLong[Int] { def toLong(value: Int) = value.toLong }
  implicit val shortToLong = new ToLong[Short] { def toLong(value: Short) = value.toLong }
}

sealed trait ValueEnum[ValueType : ToLong, EntryType <: ValueEnumEntry[ValueType]] {

  private val tl = implicitly[ToLong[ValueType]]

  final lazy val valuesToEntriesMap: LongMap[EntryType] = {
    val elems = values.map(v => tl.toLong(v.value) -> v)
    LongMap(elems)
  }

  def withValueOpt(i: ValueType): Option[EntryType] = valuesToEntriesMap.get(tl.toLong(i))
  
  // other methods unchanged

}
```

Benefits:

* No need to use `@specialized`, which is a tricky beast
* No risk of allocating an enormous sparse array

Performance impact:

* Lookups will now include the (tiny) overhead of a `i2l` instruction, but HotSpot can probably optimise it away.

shanielh:
@cb372 Lookups will still include overhead of Boxing 😞 

I guess we just need to compare the performance of both of the ways to see what's faster. I hope I'll have time to sit on this on the weekend.

cb372:
Oh yeah, `def toLong(value: A): Long` would box unless you `@specialized` it. Bleurgh.

shanielh:
And the call to the map will also box 😞

enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/IndexedEnumMap.scala
```diff
@@ -0,0 +1,73 @@
...
+  private[this] val (arrayValues, minValue) = {
+    val indices = values.map(v => toIndex(v.value))
+
+    val minValue = indices.min
+    val maxValue = indices.max
+
+    val retVal = Array.fill[Option[EntryType]](maxValue - minValue + 1)(None)
+
+    for (value <- values) {
+      retVal(toIndex(value.value) - minValue) = Some(value)
```
cb372: This won't work if the range of values is greater than IntMax. You'll run into integer overflow issues and end up trying to access a negative index of the array.

enumeratum-core/compat/src/main/scala-2.11/enumeratum/values/IndexedEnumMap.scala
```diff
@@ -0,0 +1,73 @@
...
+sealed class IntIndexedEnumMap[A <: ValueEnumEntry[Int]](values: Seq[A]) extends IndexedEnumMap[Int, A](values) {
+  final override def toIndex(value: Int): Int = value
+}
+
+sealed class ShortIndexedEnumMap[A <: ValueEnumEntry[Short]](values: Seq[A]) extends IndexedEnumMap[Short, A](values) {
+  final override def toIndex(value: Short): Int = value.toInt
+}
+
+sealed class LongIndexedEnumMap[A <: ValueEnumEntry[Long]](values: Seq[A]) extends IndexedEnumMap[Long, A](values) {
+  final override def toIndex(value: Long): Int = value.toInt
```
cb372: What if the value is < -IntMax or > IntMax?

shanielh: You're right, But this is an opt-in feature, We can add some exception in runtime that tell people that they can't use it because they're enums value range is too big.

lloydmeta:
Sorry for the delayed response. I've added benchmarks to this project (JMH) and got the following:

## Before

```
[info] Benchmark                                            Mode  Cnt     Score    Error  Units
[info] values.ValueEnumBenchmarks.withValueDoesNotExist     avgt   30  1707.968 ± 31.495  ns/op
[info] values.ValueEnumBenchmarks.withValueExists           avgt   30     3.941 ±  0.033  ns/op
[info] values.ValueEnumBenchmarks.withValueOptDoesNotExist  avgt   30     5.435 ±  0.033  ns/op
[info] values.ValueEnumBenchmarks.withValueOptExists        avgt   30     6.313 ±  0.032  ns/op
```

## After 

```
[info] Benchmark                                            Mode  Cnt     Score    Error  Units
[info] values.ValueEnumBenchmarks.withValueDoesNotExist     avgt   30  1679.596 ± 78.947  ns/op
[info] values.ValueEnumBenchmarks.withValueExists           avgt   30     9.227 ±  0.065  ns/op
[info] values.ValueEnumBenchmarks.withValueOptDoesNotExist  avgt   30     5.554 ±  0.039  ns/op
[info] values.ValueEnumBenchmarks.withValueOptExists        avgt   30     6.453 ±  0.037  ns/op
```

It's entirely possible that my benchmarks are flawed (see [here](https://github.com/lloydmeta/enumeratum/blob/master/benchmarking/compat/src/main/scala-2.11/enumeratum/values/ValueEnumBenchmarks.scala)), but it doesn't look like the changes here have improved performance in `ValueEnum`s.

lloydmeta:
Closing after many years of inactivity.

PR #43, opened 2016-07-11
Type-safe and readable contains checking

akitaylor:
In my practice I experienced that it is a usual need to check whether an enum is contained by a list of enums. 
One way is Seq(E1,E2,..).contains(E) but i think the following syntax is more readable, intuitive and type-checked:
E.in(E1,E2,...).

Please consider if this approach is right. Thanks!

### DIFF
--- a/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
@@ -64,7 +64,7 @@ object EnumEntry {
   /**
    * Helper implicit for more readable checking.
    */
-  implicit class ComparableEnum[A <: EnumEntry](enum: A) {
+  implicit class ComparableEnum[A <: EnumEntry](val enum: A) extends AnyVal {
     /**
      * Checks if the current enum value is contained by the set of enum values in the parameter list.
      * @param firstEnum First enum of the list.
--- a/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
@@ -61,4 +61,17 @@ object EnumEntry {
     abstract override def entryName: String = super.entryName.toLowerCase
   }
 
+  /**
+   * Helper implicit for more readable checking.
+   */
+  implicit class ComparableEnum[A <: EnumEntry](enum: A) {
+    /**
+     * Checks if the current enum value is contained by the set of enum values in the parameter list.
+     * @param firstEnum First enum of the list.
+     * @param otherEnums Remaining enums.
+     * @return `true` if the current value is contained by the parameter list.
+     */
+    def in(firstEnum: A, otherEnums: A*) = (firstEnum +: otherEnums).contains(enum)
+  }
+
 }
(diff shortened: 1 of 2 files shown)


lloydmeta:
@akitaylor,

Thanks for your PR. I think it's a very clever way of doing a type-checked `contains`, and the implementation looks good to me.

Can you add a test to verify that it works as expected? I would also like to see it in action, and I'm particularly interested in seeing if this requires any additional imports at the call-site (can't quite remember..) :)

akitaylor:
The tests are complete :)

Fortunately no import is needed because of the placement of the implicit class (in the EnumValue object).

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
```diff
@@ -61,4 +61,17 @@ object EnumEntry {
     abstract override def entryName: String = super.entryName.toLowerCase
   }
 
+  /**
+   * Helper implicit for more readable checking.
+   */
+  implicit class ComparableEnum[A <: EnumEntry](enum: A) {
```
lloydmeta: We should probably extend from `AnyVal` to [avoid runtime overhead](http://docs.scala-lang.org/overviews/core/value-classes.html).

enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala
```diff
@@ -250,7 +250,28 @@ class EnumSpec extends FunSpec with Matchers {
...
-        """ shouldNot compile
+      """ shouldNot compile
+    }
+  }
+
+  describe("in") {
+    import DummyEnum._
+
+    it("should return true if enum value is contained by the parameter list") {
+      val enum: DummyEnum = Hello
```
lloydmeta: Is the type ascription here necessary? I'm guessing it might be ..

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
akitaylor: The purpose of this type constraint is to restrict the implicit class to apply only to EnumEntry subtypes. I think it has only role in compile time, but it seems to be necessary.

enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala
akitaylor: Yes it is necessary, because the scala compiler is quite smart :) If you don't specify the type of "val enum:DummyEnum", then the type will be Hello.type, and the "in" function will only accept this type. But this is not a realistic scenario, in a real world situation the type of the variable will be eg. DummyEnum, not the concrete EnumValue.

lloydmeta: Right, something similar happens with integrations with [libs like Slick](https://github.com/lloydmeta/enumeratum#slick-integration). Not a problem :+1:

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
lloydmeta: Ah, maybe I should have been more clear. I wasn't talking about the type constraint; moreso the implicit class.

By having the implicit class `extend AnyVal`, we can prevent a runtime allocation because the compiler will rewrite the code into a method call on a static object.

akitaylor: Yes, it is clear now. Using value class is a good idea, indeed. I modified the code accordingly.

lloydmeta:
Thanks ! LGTM.

lloydmeta:
I've just pushed 1.4.5 to Sonatype; it should be on Maven Central shortly.

PR #47, opened 2016-07-18
Adds support for lower/uppercase transformations

mdedetrich:
This pull request adds support for lower/upper case transformations. Unlike case sensitive, these transformations will **only** support lower or upper case. Futhermore, they also alter any decoders (or writers/unbind depending on the framework that is supporting them).

This was done because in my personal use case, this logic is defined all the time. As an example we have an ENUM that is represented in a database with uppercase enum values, however the REST interface (which is implemented in play) only has lowercase names (which means we have to handle bind and unbind  transforming the value, instead of just returning the original value). Same deal with Json formatters and such

### DIFF
--- a/enumeratum-circe/src/main/scala/enumeratum/Circe.scala
+++ b/enumeratum-circe/src/main/scala/enumeratum/Circe.scala
@@ -18,12 +18,40 @@ object Circe {
     final def apply(a: A): Json = stringEncoder.apply(a.entryName)
   }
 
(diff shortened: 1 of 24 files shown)


lloydmeta:
Hi there, thanks for this PR. I agree with your use-case and the implementation is flawless :)

lloydmeta:
I've just pushed up 1.4.6 to Sonatype and it should be on Maven Central shortly. It took a while because I wanted to refactor some tests and in doing so caught a bug :). Sorry for the wait.

mdedetrich:
@lloydmeta All good, thanks for the quick release!

PR #49, opened 2016-07-20
Make format/s non private

mdedetrich:
This makes the various `format` methods non private. The reasoning is that, if you are unable to make your enums extend `PlayEnum` or its variants, the only way to create a mapping using an enum is to make a reference to it with the `enum` method, i.e.

```scala
val myEnumMapping = enumeratum.Forms.enum(MyEnum)
```

Unfortunately the only way you can use this mapping is by reference to a variable, not its type

```scala
Form(
    mapping(
        "field" -> myEnumMapping
    )
)
```

This is because play doesn't have any way to bring an implicit `Mapping` into scope, only an implicit `Formatter`. Making the format non privates means that you can do this

```scala
object Formats {
    implicit val myEnumFormat = enumeratum.Forms.format(MyEnum)
}
```

And then you can use plays `of` to summon an implicit `Format` from scope

```scala
import Formats._
Form(
    mapping(
        "field" -> of[MyEnum]
    )
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,34 @@ object GreetingForm {
 }
 ```
 
+Another alternative (if your `Enum` cant extend `PlayEnum` or `PlayFormFieldEnum`) is to create an implicit `Format`
+and bring it into scope using Play's `of`, i.e.
+
+```scala
+import play.api.data.Form
+import play.api.data.Forms._
+
+object Formats {
+  implicit val greetingFormat = enumeratum.Forms.format(Greeting)
+}
+
+object GreetingForm {
+  import Formats._
+  
+  val form = Form(
+      mapping(
+        "name" -> nonEmptyText,
+        "greeting" -> of[Greeting]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -354,6 +354,34 @@ object GreetingForm {
 }
 ```
 
+Another alternative (if for example your `Enum` cant extend `PlayEnum` or `PlayFormFieldEnum`) is to create an implicit `Format`
```
lloydmeta: Small typo: cant → can't

lloydmeta:
Thanks this sounds reasonable, just have a small typo in there that needs fixing and we're good to go :+1:

README.md
mdedetrich: Fixed

lloydmeta:
Thanks. 1.4.8 should in the next hour or so be in Maven Central.

PR #52, opened 2016-07-24
Infer Enum[A] from A <: EnumEntry

rpiaggio:
Added a macro to infer automatically the Enum object of an EnumEntry.

This can be used to simplify mappings, serialization, etc.. For example, we can now do:

* An automatic Play JSON Formats:
```
implicit def enumFormat[A <: EnumEntry : Enum]: Format[A] = EnumFormats.formats(implicitly[Enum[A]])
```

* An automatic Slick mapper:

```
  implicit def enumMapper[A <: EnumEntry : Enum : ClassTag] = MappedColumnType.base[A, String](
    sn => sn.entryName,
    str => implicitly[Enum[A]].withName(str)
  )
```

* An io.underscore Validator:

```
  implicit def enumValidator2stringValidator[A <: EnumEntry : Enum](enumValidator: Validator[A]): Validator[String] =
    Validator[String] { in => implicitly[Enum[A]].withNameOption(in) map enumValidator getOrElse fail("Value is required") }
```

Etc...

### DIFF
--- a/enumeratum-core/src/main/scala/enumeratum/Enum.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/Enum.scala
@@ -2,6 +2,7 @@ package enumeratum
 
 import scala.language.experimental.macros
 import scala.language.postfixOps
+import ContextUtils.Context
 
 /**
  * All the cool kids have their own Enumeration implementation, most of which try to
@@ -149,4 +150,19 @@ trait Enum[A <: EnumEntry] {
 
   private lazy val existingEntriesString = values.map(_.entryName).mkString(", ")
 
+}
+
+object Enum {
+  implicit def materializeEnum[A <: EnumEntry]: Enum[A] = macro EnumMacrosAux.materializeEnumImpl[A]
+}
+
+object EnumMacrosAux {
+  /**
+   * Given an A <: EnumEntry, provide Enum[A]
+   */
+  def materializeEnumImpl[A <: EnumEntry: c.WeakTypeTag](c: Context) = {
+    import c.universe._
+    val symbol = weakTypeOf[A].typeSymbol
+    c.Expr[Enum[A]](Ident(symbol.companionSymbol))
+  }
 }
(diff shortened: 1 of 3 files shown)


rpiaggio:
Sorry about the test not passing, will fix that and add tests for the new functionality...

rpiaggio:
@lloydmeta There seems to be a problem with using quasiquotes in Scala 2.10. Macro paradise seems to be imported (https://github.com/lloydmeta/enumeratum/blob/master/project/Build.scala#L137) but it's not working anyway. Should I avoid using quasiquotes altogether?

lloydmeta:
Hi there,

Thanks for your interest and the cool PR. Definitely looks like a useful thing to have!

And yeah, the root project (well, the macro one) doesn't have a dependency on Macro Paradise because I wanted ppl using the project to have as few dependencies as possible. What you're seeing is the subproject for UPickle integration.

So yes, It would be great if you could implement the macro without using Quasiquotes !

rpiaggio:
Actually thank you for this great project.

I made a new commit. It is working without the quasiquote now, sorry I didn't notice Marcro Paradise was being used only in the UPickle subproject.

I also added a test.

rpiaggio:
Also, please feel free to relocate the implicit def to wherever you see fit.

rpiaggio:
Ooops... I actually tested with `+test` instead of `+test:compile`.

I now fixed the problem, but the only way I could find is to declare a macro in the core project, since the macro project doesn't know the `Enum` type.

rpiaggio:
I can squash the whole thing into a single commit if you prefer.

lloydmeta:
Nice, you got it to work without quasiquotes ! I may try to move the macro around before releasing later today, but this looks great to me. Thanks !

lloydmeta:
Just released 1.4.8. It should be up on Maven Central now.

rpiaggio:
Thank you very much! I started working on removing the enumeratum dependencies from the macro and moved it back to the macros project, but realized you already did that and much more.

Thanks again!

PR #72, opened 2016-09-22
Documentation on how to use slick interpolated queries.

bomgar:
Add an example to the documentation.


lloydmeta:
Thanks a lot for fleshing out the Slick documentation :+1:

PR #76, opened 2016-10-14
Add support for Argonaut

alonsodomin:
Support for encoding and decoding values using Argonaut.

### DIFF
--- a/enumeratum-argonaut/src/main/scala/enumeratum/values/ArgonautValueEnum.scala
+++ b/enumeratum-argonaut/src/main/scala/enumeratum/values/ArgonautValueEnum.scala
@@ -1,0 +1,75 @@
+package enumeratum.values
+
+import argonaut._
+
+/**
+  * Created by alonsodomin on 14/10/2016.
+  */
+sealed trait ArgonautValueEnum[ValueType, EntryType <: ValueEnumEntry[ValueType]] {
+  this: ValueEnum[ValueType, EntryType] =>
+
+  implicit def argonautEncoder: EncodeJson[EntryType]
+
+  implicit def argonautDecoder: DecodeJson[EntryType]
+
+}
+
+/**
+  * ArgonautEnum for IntEnumEntry
+  */
+trait IntArgonautEnum[EntryType <: IntEnumEntry] extends ArgonautValueEnum[Int, EntryType] {
+  this: ValueEnum[Int, EntryType] =>
+
+  implicit val argonautEncoder = Argonauter.encoder(this)
+  implicit val argonautDecoder = Argonauter.decoder(this)
+}
+
+/**
+  * ArgonautEnum for LongEnumEntry
+  */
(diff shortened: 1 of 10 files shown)


enumeratum-argonaut/src/main/scala/enumeratum/values/ArgonautValueEnum.scala
```diff
@@ -0,0 +1,75 @@
...
+  implicit val argonautDecoder = Argonauter.decoder(this)
+}
+
+/**
+  * ArgonautEnum for ByteEnumEntry
+  */
+trait ByteArgonautEnum[EntryType <: ByteEnumEntry] extends ArgonautValueEnum[Byte, EntryType] {
+  this: ValueEnum[Byte, EntryType] =>
+
+  implicit val argonautEncoder = Argonauter.encoder(this)
```
lloydmeta: [Looks like](https://travis-ci.org/lloydmeta/enumeratum/builds/167795213#L923) the compiler can't find the `EncodeJson[Byte]` implicit declared inside the package object. I'm not exactly sure why; could be something to do with the fact that this file comes first alphabetically ? 😨

enumeratum-argonaut/src/test/scala/enumeratum/TrafficLight.scala
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+  * Created by alonsodomin on 14/10/2016.
+  */
+sealed trait TrafficLight extends EnumEntry
+object TrafficLight extends Enum[TrafficLight] with ArgonautEnum[TrafficLight] {
+  case object Red    extends TrafficLight
+  case object Yellow extends TrafficLight
+  case object Green  extends TrafficLight
+  override def values: Seq[TrafficLight] = Seq(Red, Yellow, Green)
```
lloydmeta: `val values = findValues` might be nicer here.

lloydmeta:
Thanks for the PR. I especially liked your use of `Short` when decoding `Byte`; I should probably replace that elsewhere too :)

Left a bit of feedback here and there, no rush.

alonsodomin:
That compile error due to missed implicits was a weird one! I could only reproduce it by compiling from scratch in 2.10... the fix apparently is to add type annotations to the encoder/decoder, after that it can resolve the implicit (I finally added the type annotations to all the enum values, for consistency).

btw, for some reason scalafmt was making SBT fail during compile time when working on 2.10, are you familiar with that? I temporarly disabled the formatting to work on this, hence the last "futile" commit to revert it back.

lloydmeta:
@alonsodomin strange, I don't seem to be having issues with scalafmt when working on 2.10.

In any case, everything looks good now ! Thanks a lot !

lloydmeta:
Just pushed `enumeratum-argonaut`, version `1.4.15` to Sonatype releases. It should be up on Maven central in the next hour or so. Thanks again 👍

alonsodomin:
Didn't expect such a quick turnover but that's great, it will help removing some boilerplate in clien projects.

Issue #84, opened 2016-12-01
Enum.withName safe alternative


pedrorijo91:
Enum.withName throws an exception for invalid names. Would it be possible to provide a method that would simply return an option/try/something else instead? maybe even deprecate the current method

lloydmeta:
I agree. The reason `withName` throws when there is no corresponding entry is to follow the same behaviour as standard lib's `Enumeration` API. That way migration is easier.

That said there is a `withNameOption` method for `Enum` that does what you want (see [here](https://github.com/lloydmeta/enumeratum/blob/master/enumeratum-core/src/main/scala/enumeratum/Enum.scala#L87)). Not only is the API nicer and safer as you point out, but it is also close to 300x faster when there is no such entry (see [benchmarks](https://github.com/lloydmeta/enumeratum#benchmarking)), likely because there is no need to build a stacktrace.

pedrorijo91:
oh sorry, didn't see that method. maybe we could just refer in the README the alternative? it's not very visible in the README imho

lloydmeta:
That sounds like a good idea. If you send a PR I'd be glad to add it to the Readme :)

pedrorijo91:
sure, i'll take care of it today asap :)

lloydmeta:
Fixed via #85

PR #85, opened 2016-12-01
Add Enum.withNameOption example usage to README

pedrorijo91:
Solves #84 by making it more explicit the existence of a safe method to get an instance of Enum by its name


lloydmeta:
Perfect thanks !

PR #123, opened 2017-03-06
Json4s integration, fixes #121

### LINKED ISSUE #121: Integration with Json4s
Would it be possible to have an integration with [Json4s](http://json4s.org/)? Serializing an enum is trivial:

```scala
import enumeratum.EnumEntry
import org.json4s.CustomSerializer

object EnumSerializer extends CustomSerializer[EnumEntry](format => (
    PartialFunction.empty,
    {
      case x: EnumEntry => JString(x.entryName)
    }
  ))
```

But deserializing is not so easy, the tricky part is to instantiate the correct enum type.

mikaelstaldal:
Fixes issue #121

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,8 @@ lazy val integrationProjectRefs = Seq(
   enumeratumCirceJs,
   enumeratumCirceJvm,
   enumeratumReactiveMongoBson,
-  enumeratumArgonaut
+  enumeratumArgonaut,
+  enumeratumJson4s
 ).map(Project.projectToRef)
 
 lazy val root =
@@ -77,6 +78,7 @@ lazy val scala_2_12 = Project(id = "scala_2_12",
         enumeratumUPickleJs,
         enumeratumUPickleJvm,
         enumeratumArgonaut,
+        enumeratumJson4s,
         enumeratumReactiveMongoBson
       ).map(Project.projectToRef): _*) // base plus known 2.12 friendly libs
 
--- a/build.sbt
+++ b/build.sbt
@@ -16,6 +16,7 @@ lazy val reactiveMongoVersion = "0.12.1"
 lazy val circeVersion         = "0.7.0"
 lazy val uPickleVersion       = "0.4.4"
 lazy val argonautVersion      = "6.2-RC2"
+lazy val json4sVersion        = "3.5.0"
 def thePlayVersion(scalaVersion: String) =
   CrossVersion.partialVersion(scalaVersion) match {
     case Some((2, scalaMajor)) if scalaMajor >= 11 => "2.5.12"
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -275,6 +276,21 @@ lazy val enumeratumArgonaut =
       )
     )
 
+lazy val enumeratumJson4s =
```
lloydmeta: Can you please add references to this project 

1. [In this list of integration projects](https://github.com/lloydmeta/enumeratum/blob/master/build.sbt#L38)
2. [In this list of 2.12-compatible projects](https://github.com/lloydmeta/enumeratum/blob/master/build.sbt#L79) (I think I saw a 2.12 artefact)

It's a bit of a hassle but we're still waiting on Play to come out with their 2.12 release :)

enumeratum-json4s/src/main/scala/enumeratum/values/Json4s.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.json4s.CustomSerializer
+import org.json4s.JsonAST.{JInt, JLong, JString}
+
+
+object Json4s {
+
+  def serializer[A <: IntEnumEntry: Manifest](enum: IntEnum[A]): CustomSerializer[A] = new CustomSerializer[A](_ => (
+    {
+      case JInt(i) if enum.withValueOpt(i.toInt).isDefined => enum.withValue(i.toInt)
+      case JLong(i) if enum.withValueOpt(i.toInt).isDefined => enum.withValue(i.toInt)
```
lloydmeta: I'm a bit hesitant about doing this, because we could get false matches.

```scala
@ Long.MaxValue 
res0: Long = 9223372036854775807L
@ Long.MaxValue.toInt 
res1: Int = -1
```

Actually, I just realised that `JInt` is a `BigDecimal` (see [here](https://github.com/json4s/json4s/blob/0b28498f065ed8365e4e336a5fec1ec5d4e15649/ast/src/main/scala/org/json4s/JsonAST.scala#L145)), so maybe we should be doing bounds checks for JInt too...

enumeratum-json4s/src/main/scala/enumeratum/values/Json4s.scala
```diff
@@ -0,0 +1,67 @@
...
+      case JLong(i) if enum.withValueOpt(i).isDefined => enum.withValue(i)
+    },
+    {
+      case x: A => JLong(x.value)
+    }
+  ))
+
+  def serializer[A <: ShortEnumEntry: Manifest](enum: ShortEnum[A]): CustomSerializer[A] = new CustomSerializer[A](_ => (
+    {
+      case JInt(i) if enum.withValueOpt(i.toShort).isDefined => enum.withValue(i.toShort)
```
lloydmeta: Here too, we should have guards in case the numbers are too big to fit inside the Short range

enumeratum-json4s/src/main/scala/enumeratum/values/Json4s.scala
```diff
@@ -0,0 +1,67 @@
...
+      case JString(s) if enum.withValueOpt(s).isDefined => enum.withValue(s)
+    },
+    {
+      case x: A => JString(x.value)
+    }
+  ))
+
+  def serializer[A <: ByteEnumEntry: Manifest](enum: ByteEnum[A]): CustomSerializer[A] = new CustomSerializer[A](_ => (
+    {
+      case JInt(i) if enum.withValueOpt(i.toByte).isDefined => enum.withValue(i.toByte)
```
lloydmeta: Byte bound check might make sense.

build.sbt
mikaelstaldal: Fixed.

mikaelstaldal:
@lloydmeta I think I have addressed all issues now.

lloydmeta:
Released as `enumeratum-json4s` on [Sonatype](https://oss.sonatype.org/#nexus-search;quick~enumeratum-json4s). Should be on Maven Central within an hour or so.

PR #127, opened 2017-03-16
Implement Json4s CustomKeySerializer

morgen-peschke:
I ran into an issue where an enum was not deserializing in the expected manner using Json4s. After a bit of debugging, it turns out Json4s uses a separate serializer to convert keys.

The `CustomKeySerializer` class is nearly identical to the `CustomSerializer` class, so it creating a new helper was fairly straightforward, once it's need became apparent.

### DIFF
--- a/enumeratum-json4s/src/main/scala/enumeratum/Json4s.scala
+++ b/enumeratum-json4s/src/main/scala/enumeratum/Json4s.scala
@@ -1,6 +1,6 @@
 package enumeratum
 
-import org.json4s.CustomSerializer
+import org.json4s.{CustomSerializer, CustomKeySerializer}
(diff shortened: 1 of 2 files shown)


lloydmeta:
This looks great. I just did a quick check of the json4s.org website and realised that they don't mention this class at all, so thanks for figuring that out.

morgen-peschke:
Any time :)

lloydmeta:
I just released enumeratum-json4s v1.5.10 on Sonatype. It should be on Maven Central within the next hour or so.

morgen-peschke:
I've pulled it, and everything looks good. Thanks for the quick turnaround 👍

PR #133, opened 2017-04-05
[WIP] Upgrade/play 2.6.0 m

lloydmeta:
Waiting on:
- [ ] Play 2.6.0 official release
  - [ ] Remove suffix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: scala
 
-# Only run aggregation 2.11, for 2.12, only run the special dummy project
+# Only run aggregation 2.11
 matrix:
   include:
     - scala: 2.10.6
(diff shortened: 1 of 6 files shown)


ShaneDelmore:
I have been using this branch with no problems, thank you for publishing it.  If you update it to Play 2.6-m7 would you consider publishing this as a -M branch instead of -SNAPSHOT?  It's a small difference but being a snapshot invalidates all sbt caching when using it as it is always checking if a newer snapshot has been published which I would not expect being based on a milestone release.

lloydmeta:
@ShaneDelmore good point; done. I just published new versions of the 2.6.0 integrations:

- `enumeratum-play` (based on M5, which is the latest according to [the official site](https://www.playframework.com/download#alternatives))
- `enumeratum-play-json` (based on M7)

both prefixed with `2.6.0-Mx`

PR #136, opened 2017-04-19
Implement stackable Camelcase, LowerCamelcase and Uncapitalised

### LINKED ISSUE #125: Camelcase mixin
I've noticed there's an `EnumEntry` mixin for pretty much any case I could want, except camelCase, which is by far the most common case I deal with.

dmit:
Closes #125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ State.withName("AL")
 #### Mixins to override the name
 
 The second way to override the name behaviour is to mixin the stackable traits provided for common string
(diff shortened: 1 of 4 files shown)


dmit:
The 2.10 build failed (twice) due to
wartremover/wartremover#127 which points to
wartremover/wartremover#74 which points to
scala/bug#6240 that was fixed in 2.11, but not backported.

lloydmeta:
Thanks this looks good to me. Also thanks for tracking down that weird intermittent error; it had been bothering me but I never got around to figuring it out :)

lloydmeta:
Just released this as part of 1.5.12.

PR #137, opened 2017-04-19
1.5.12 release

lloydmeta:


build.sbt
```diff
@@ -1,12 +1,12 @@
 import com.typesafe.sbt.SbtGit.{GitKeys => git}
 
-lazy val theScalaVersion = "2.11.8"
+lazy val theScalaVersion = "2.11.10"
```
ShaneDelmore: 2.11.10 had a binary compat issue. Use 2.11.11 instead.

lloydmeta: Thanks for the heads up. Will fix.

PR #150, opened 2017-08-24
Balance single quotes in decoding failure message

tbl3rd:
I noticed some strange single-quotes in some test runs here.
And thank you for enumeratum.


lloydmeta:
Thanks for fixing this !

tbl3rd:
We might even want a space before the quoted term.

PR #154, opened 2017-10-23
Expands macros in ValueEnumMacros

jackiechun:
I ran into an issue with ValueEnumMacros, I want to write something like this with binary notation. It was important for me to make sure all of them are different and form an enumeration, but seeing the binary values is important. I don't want to write them as base 10 integers.

```scala
sealed abstract class MacroEval(val value: Int) extends IntEnumEntry

case object MacroEval extends IntEnum[MacroEval] {
  import spire.syntax.literals.radix._

  case object Zero   extends MacroEval(x2"0")
  case object One    extends MacroEval(x2"1")
  case object Two    extends MacroEval(x2"10")

  val values = findValues
}
```

The `x2"010101"` notation is done with macros from Spire and expands into a integer at compile time. Except it won't work with Enumeratum as is because the macros for ValueEnum won't evaluate the expression they receive and rather expect a Literal.

I implemented something that seems to not break stuff and work for my use case. I had to mess with the SBT build a bit. I don't understand why the sub projects depend on external binaries and not the code. It would be nice to have some pointer on how to develop on the project.

Anyway, I wanted to post something that works and can be tested. Let me know what you think. Before merging I will revert those changes off course.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -97,11 +97,12 @@ lazy val core = crossProject
   .settings(
     name := "enumeratum",
     version := Versions.Core.head,
-    libraryDependencies += "com.beachape" %% "enumeratum-macros" % Versions.Macros.stable
+    crossScalaVersions := scalaVersions
+//    libraryDependencies += "com.beachape" %% "enumeratum-macros" % Versions.Macros.stable
   )
   .settings(testSettings: _*)
   .settings(commonWithPublishSettings: _*)
-//  .dependsOn(macros) // used for testing macros
+  .dependsOn(macros) // used for testing macros
 lazy val coreJS  = core.js
 lazy val coreJVM = core.jvm
 
@@ -114,18 +115,11 @@ lazy val enumeratumTest = crossProject
   .settings(commonWithPublishSettings: _*)
   .settings(
     name := "enumeratum-test",
-    version := Versions.Core.stable,
-    libraryDependencies += {
-      import org.scalajs.sbtplugin._
-      val crossVersion =
-        if (ScalaJSPlugin.autoImport.jsDependencies.?.value.isDefined)
-          ScalaJSCrossVersion.binary
-        else
-          CrossVersion.binary
-      impl.ScalaJSGroupID
-        .withCross("com.beachape", "enumeratum", crossVersion) % Versions.Core.stable
-    }
+    version := Versions.Core.head,
+    crossScalaVersions := scalaVersions,
+    libraryDependencies += "org.spire-math" %% "spire" % "0.13.0"
   )
+  .dependsOn(core)
 lazy val enumeratumTestJs  = enumeratumTest.js
 lazy val enumeratumTestJvm = enumeratumTest.jvm
 
@@ -151,7 +145,7 @@ lazy val enumeratumReactiveMongoBson =
     .settings(testSettings: _*)
     .settings(
       crossScalaVersions := scalaVersions,
-      version := "1.5.13-SNAPSHOT",
+      version := Versions.Core.head,
       libraryDependencies ++= Seq(
         "org.reactivemongo" %% "reactivemongo"   % reactiveMongoVersion,
         "com.beachape"      %% "enumeratum"      % Versions.Core.stable,
@@ -164,7 +158,7 @@ lazy val enumeratumPlayJson = Project(id = "enumeratum-play-json",
                                       settings = commonWithPublishSettings)
   .settings(testSettings: _*)
   .settings(
-    version := s"1.5.13-SNAPSHOT",
+    version := Versions.Core.head,
     crossScalaVersions := scalaVersions,
     libraryDependencies ++= Seq(
       "com.typesafe.play" %% "play-json"       % thePlayJsonVersion(scalaVersion.value),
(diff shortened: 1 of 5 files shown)


macros/compat/src/main/scala-2.11/enumeratum/ContextUtils.scala
```diff
@@ -37,4 +37,10 @@ object ContextUtils {
   def constructorName(c: Context): c.universe.TermName = {
     c.universe.termNames.CONSTRUCTOR
   }
+
+  def evalTree[ValueType](c: Context)(expr: c.universe.Tree): ValueType = {
+    @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements")) // false positive bug in WartRemover?
+    val evaluated = c.eval(c.Expr[ValueType](c.untypecheck(c.typecheck(expr))))
```
jackiechun: `eval` will only evaluate untyped expressions. For some reason it says `expr` is typechecked and refuses to work. Performing an `untypecheck` will complain that `expr` is actually not typed so it can't untype it.... So we run `typecheck` to make it very typed and then `untypecheck` it so it's not typed and `eval` likes it.

I wasted days on this...

macros/src/main/scala/enumeratum/ValueEnumMacros.scala
```diff
@@ -217,6 +217,14 @@ object ValueEnumMacros {
                       c.enclosingPosition,
                       s"${declTree.symbol} has a value with the wrong type: $i:${i.getClass}, instead of ${classTag.runtimeClass}"
                     )
+
+                  case (_, AssignOrNamedArg(Ident(`valueTerm`), expr)) =>
```
jackiechun: I think the other cases could be refactored using `expr` en `eval` and remove some code.

macros/compat/src/main/scala-2.10/enumeratum/ContextUtils.scala
```diff
@@ -37,4 +37,10 @@ object ContextUtils {
   def constructorName(c: Context): c.universe.TermName = {
     c.universe.nme.CONSTRUCTOR
   }
+
+  def evalTree[ValueType](c: Context)(expr: c.universe.Tree): ValueType = {
+    @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements")) // false positive bug in WartRemover?
+    val evaluated = c.eval(c.Expr[ValueType](expr))
```
jackiechun: Scala 2.10 doesn't support typecheck and untypecheck but it seems to just work anyway.

lloydmeta:
Hi there, 

Thanks for this contribution! It looks great at a glance; I'll go over it in more detail by the end of the week, but I just wanted to shed some light on:

>  I had to mess with the SBT build a bit. I don't understand why the sub projects depend on external binaries and not the code. It would be nice to have some pointer on how to develop on the project.

The reason for this is because we decouple the versioning across the different layers, so that they can be released on separate schedules. Further, I wanted to make sure that the higher layers (e.g. integration libs) would work with the latest released versions of the inner-layers (e.g. macro, core). At least that was the aim. I'm definitely open to re-jigging the build to make it less annoying to develop this lib!!

At the moment, in order to test the macros code in, say core, I add a small line like [this](https://github.com/lloydmeta/enumeratum/blob/master/build.sbt#L104).

build.sbt
```diff
@@ -164,7 +158,7 @@ lazy val enumeratumPlayJson = Project(id = "enumeratum-play-json",
                                       settings = commonWithPublishSettings)
   .settings(testSettings: _*)
   .settings(
-    version := s"1.5.13-SNAPSHOT",
+    version := Versions.Core.head,
```
lloydmeta: I think a bunch of these `version := Versions.Core.head` can be reverted as you've mentioned.

enumeratum-test/src/main/scala/enumeratum/values/MacroEval.scala
```diff
@@ -0,0 +1,14 @@
+package enumeratum.values
+
+sealed abstract class MacroEval(val value: Int) extends IntEnumEntry
```
lloydmeta: The `enumeratum-test` code is used to declare models that might be reused later in other subprojects (e.g. in integration libs).

Since this isn't be used for that purpose, it might be a good idea to move it over to `enumeratum-core` and write a few tests to make sure the `value`s of each enum member get passed/evaluated properly, e.g.

```scala
MacroEval.Zero.value shouldBe 0
MacroEval.One.value shouldBe 1
MacroEval.Two.value shouldBe 2
MacroEval.Eleven.value shouldBe 11
```

macros/compat/src/main/scala-2.10/enumeratum/ContextUtils.scala
lloydmeta: Yeah, the Scala macros API changed quite a bit between 2.10 and 2.11, hence the ContextUtils shim. Thanks for doing the legwork to get it working!!

macros/src/main/scala/enumeratum/ValueEnumMacros.scala
lloydmeta: Agreed. I think it might be able to get rid of the special handling per value type. Likely best to do it in another PR though :)

lloydmeta:
Closing after many years of inactivity.

PR #164, opened 2018-02-01
Feature/scalacheck integration

Sufflope:
Provide default Arbitrary and Cogen instances for all Enum types.

### DIFF
--- a/enumeratum-scalacheck/src/main/scala/enumeratum/CogenInstances.scala
+++ b/enumeratum-scalacheck/src/main/scala/enumeratum/CogenInstances.scala
@@ -1,0 +1,14 @@
+package enumeratum
+
+import org.scalacheck.Cogen
+
+trait CogenInstances {
+
+  /**
+    * `Enum` context bound is unused but can drastically reduce compilation time.
+    * See https://github.com/nrinaudo/kantan.codecs/blob/bb74def19e94ce4f14330100b467c3fc9271068d/enumeratum/core/src/main/scala/kantan/codecs/enumeratum/values/ValueEnumInstances.scala#L82
+    */
+  implicit def cogenEnumEntry[EnumType <: EnumEntry: Enum]: Cogen[EnumType] =
+    Cogen[String].contramap(_.entryName)
+
+}
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Integrations are available for:
 - [ReactiveMongo BSON](http://reactivemongo.org/releases/0.11/documentation/bson/overview.html): JVM only
 - [Argonaut](http://argonaut.io): JVM and ScalaJS
 - [Json4s](http://json4s.org): JVM only
+- [ScalaCheck](https://www.scalacheck.org): JVM only
 
 ### Table of Contents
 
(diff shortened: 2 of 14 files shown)


enumeratum-scalacheck/src/main/scala/enumeratum/CogenInstances.scala
```diff
@@ -0,0 +1,10 @@
+package enumeratum
+
+import org.scalacheck.Cogen
+
+trait CogenInstances {
+
+  implicit def cogenEnumEntry[EnumType <: EnumEntry]: Cogen[EnumType] =
```
nrinaudo: I'm putting this comment here, but it applies to all instances of `Cogen`.

I have found, when writing `enumeratum` instances for my various kantan libraries, that even if requiring an implicit `Enum` instance is not strictly necessary, it helps the compiler a *lot* when resolving implicits - without the constraint some code took entire minutes to compile, and it became instantaneous with the constraint.

I decided to add the constraint in my code because:
* it improves the experience so much
* it has no cost
* unless I'm mistaken, every sane `EnumEntry` subtype *will* always have an implicit `Enum` in the global implicit scope.

lloydmeta: That sounds like an interesting finding and if it improves the development experience, I think it's a great idea to add that constraint. 

That said, unless I'm missing something (completely possible!), it doesn't look like there is any constraint here that `EnumEntry` has an `Enum` instance (e.g. `EnumType <: EnumEntry : Enum`)?

Sufflope: That's what Nicolas is talking about (adding `: Enum` constraint to `EnumType`).

In `Cogen` signature I didn't put it because it is not actually needed here (unlike for `Arbitrary` where I actually use the `Enum` instance to get the values), but from Nicolas' experience, adding it (although we won't use it) can make compilation faster.

The downside is that we won't be able to derive `Cogen` for `EnumEntry` which has no associated `Enum`, but that would be a weird situation anyway…

So if you're OK with it I can probably add this constraint.

lloydmeta: @nrinaudo @Sufflope Ugh, apologies, I didn't even realise that @nrinaudo was not the one who issued this PR and thought he was reviewing his own PR 😅 

@Sufflope I think adding the `: Enum` constraint would be reasonable. Asides from improving the dev experience (minutes -> instant sounds amazing, though I'm now more curious to understand why??), as you said, `Cogen` for `EnumEntry` with no `Enum` is kind of strange anyway.

So yea, if it isn't too much trouble can you please add the constraint (and perhaps in the code, leave a small link to this comment thread, so nobody feels the itch to take it wawy)? 

Thanks in advance :) and once again, sorry for my mistake.

nrinaudo: No worries, it's kind of rare for random people to ninja-review PRs like that. For context, @Sufflope and I are colleagues, that's how I knew to take a look.

The minutes -> instant thing was with more complicated types - basically `(A: Enum) => Either[E, B]`. My reasoning was that adding an `Enum` constraint reduces the search space drastically - most types don't have an `Enum` instance in scope, but most types *might* be a subtype of `EnumEntry`. If the compiler starts by applying the typeclass constraint before checking the subtyping one, it kind of makes sense. Maybe. Or maybe I've no idea what I'm talking about...

Sufflope:
I added the context bounds, and commented on the rationale behind them, for future readers.

lloydmeta:
@Sufflope @nrinaudo thanks for the great work :) I've just merged and published this to Sonatype; it should be on Maven Central soon-ish.

Sufflope:
Great, thanks a lot!

Issue #163, opened 2018-02-01
Quill integration


daniel-shuy:
I have a suggestion to add a trait to automatically encode/decode `EnumEntry`s as `String`s, using their `entryName`, when querying/inserting from/into a Database using [Quill](http://getquill.io).

This is my current implementation:
```scala
import enumeratum.{Enum, EnumEntry}
import io.getquill.MappedEncoding

trait QuillEnum[A <: EnumEntry] {
  this: Enum[A] =>

  implicit lazy val enumEncoder: MappedEncoding[A, String] = MappedEncoding(_.entryName)

  implicit lazy val enumDecoder: MappedEncoding[String, A] = MappedEncoding(withName)
}
```

lloydmeta:
Sounds like a good idea :) A PR for adding said integration is welcome !

PR #170, opened 2018-02-12
Add enumeratum-quill

daniel-shuy:
As mentioned in #163.

Adds `QuillEnum` trait to automatically convert `EnumEntry`s to/from their `entryName`.
Adds `QuillValueEnum` trait to automatically convert `ValueEnumEntry`s to/from their `value`s.

I wasn't sure what to put for the `version`, so I left it as `0.1.0-SNAPSHOT`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -894,12 +894,14 @@ case class Shirt(size: ShirtSize)
 
 import io.getquill._
 
-val ctx = new SqlMirrorContext(MirrorSqlDialect, Literal)
+lazy val ctx = new PostgresJdbcContext(SnakeCase, "ctx")
 import ctx._
 
-val sql = ctx.run(query[Shirt].insert(_.size -> lift(ShirtSize.Small: ShirtSize))).string
-assert(sql == "INSERT INTO Shirt (size) VALUES (?)")
+ctx.run(query[Shirt].insert(_.size -> lift(ShirtSize.Small: ShirtSize)))
+
+ctx.run(query[Shirt]).foreach(println)
 ```
+- Note that an explicit cast to the `EnumEntry` trait (eg. `ShirtSize.Small: ShirtSize`) is required when binding hardcoded `EnumEntry`s
 
 #### ValueEnum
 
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -301,6 +302,33 @@ lazy val enumeratumScalacheck =
...
+          CrossVersion.binary
+      }
+      Seq(
+        impl.ScalaJSGroupID.withCross("io.getquill", "quill-core", cross)  % quillVersion,
+        impl.ScalaJSGroupID.withCross("io.getquill", "quill-sql", cross)   % quillVersion % Test,
+        impl.ScalaJSGroupID.withCross("com.beachape", "enumeratum", cross) % Versions.Core.stable
+      )
+    }
+  )
+lazy val enumeratumQuillJs  = enumeratumQuill.js
```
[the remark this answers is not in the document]
daniel-shuy: Haha I was shocked as well

daniel-shuy:
Just discovered that all Quill contexts don't have an Encoder/Decoder for `Char`, so I had to convert `CharEnum` to `String` instead.

daniel-shuy:
- `quill-cassandra` is missing Encoders/Decoders for `Short` and `Byte` (getquill/quill#1009), so it cannot be used with `ShortEnum`/`ByteEnum`
- `quill-orientdb` is missing Encoders/Decoders for `Byte` (getquill/quill#1029), so it cannot be used with `ByteEnum`

I've added a note on them in the `README`

README.md
```diff
@@ -848,6 +850,91 @@ Similarly, you can get `Arbitrary` and `Cogen` instances for every `ValueEnum` s
...
+import io.getquill._
+
+lazy val ctx = new PostgresJdbcContext(SnakeCase, "ctx")
+import ctx._
+
+ctx.run(query[Shirt].insert(_.size -> lift(ShirtSize.Small: ShirtSize)))
+
+ctx.run(query[Shirt]).foreach(println)
+```
+- Note that an explicit cast to the `ValueEnumEntry` abstract class (eg. `ShirtSize.Small: ShirtSize`) is required when binding hardcoded `ValueEnumEntry`s
```
lloydmeta: No big deal; but this isn't an explicit cast; just an explicit type annotation :)

daniel-shuy: Ah, I knew that it wasn't an explicit cast (`asInstanceOf`), but I forgot what it was called, thanks

daniel-shuy: Actually, if we were to be pedantic, I guess this should be called a [Type Ascription](https://docs.scala-lang.org/style/types.html#ascription) instead of a Type Annotation?

daniel-shuy: Also, you missed out on one of the "explicit cast"s. I've opened a PR (#172) to fix both.

daniel-shuy:
Oops, I forgot to run `test` with crossScalaVersion (`+ test`). I'm surprised the CI passed O_o

PR #172, opened 2018-02-14
Update Quill Integration notes in README

daniel-shuy:
Should be a [Type Ascription](https://docs.scala-lang.org/style/types.html#ascription) instead of an explicit cast or Type Annotation.


lloydmeta:
Thanks!

PR #175, opened 2018-02-19
Many to one

neilchaudhuri:
Adds support for `MultiEnum`, which maps multiple entries to a single enum. Also uses `breakOut` for some minor memory optimization.

### DIFF
--- a/enumeratum-core/src/main/scala/enumeratum/Enum.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/Enum.scala
@@ -3,6 +3,7 @@ package enumeratum
 import scala.language.experimental.macros
 
 import scala.collection.immutable._
+import scala.collection.breakOut
 
 /**
   * All the cool kids have their own Enumeration implementation, most of which try to
@@ -38,20 +39,20 @@ trait Enum[A <: EnumEntry] {
   /**
     * Map of [[A]] object names to [[A]]s
     */
-  lazy final val namesToValuesMap: Map[String, A] =
-    values.map(v => v.entryName -> v).toMap
+  lazy val namesToValuesMap: Map[String, A] =
+    values.map(v => v.entryName -> v)(breakOut)
 
   /**
     * Map of [[A]] object names in lower case to [[A]]s for case-insensitive comparison
     */
   lazy final val lowerCaseNamesToValuesMap: Map[String, A] =
-    values.map(v => v.entryName.toLowerCase -> v).toMap
+    values.map(v => v.entryName.toLowerCase -> v)(breakOut)
 
   /**
     * Map of [[A]] object names in upper case to [[A]]s for case-insensitive comparison
     */
   lazy final val upperCaseNameValuesToMap: Map[String, A] =
-    values.map(v => v.entryName.toUpperCase -> v).toMap
+    values.map(v => v.entryName.toUpperCase -> v)(breakOut)
 
   /**
     * Map of [[A]] to their index in the values sequence.
--- a/enumeratum-core/src/main/scala/enumeratum/Enum.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/Enum.scala
@@ -44,13 +44,13 @@ trait Enum[A <: EnumEntry] {
     * Map of [[A]] object names in lower case to [[A]]s for case-insensitive comparison
     */
   lazy final val lowerCaseNamesToValuesMap: Map[String, A] =
-    values.map(v => v.entryName.toLowerCase -> v).toMap
+    namesToValuesMap.map{ case (k, v) => k.toLowerCase -> v }
 
   /**
     * Map of [[A]] object names in upper case to [[A]]s for case-insensitive comparison
     */
   lazy final val upperCaseNameValuesToMap: Map[String, A] =
-    values.map(v => v.entryName.toUpperCase -> v).toMap
+    namesToValuesMap.map{ case (k, v) => k.toUpperCase() -> v }
 
   /**
     * Map of [[A]] to their index in the values sequence.
--- a/enumeratum-core/src/main/scala/enumeratum/Enum.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/Enum.scala
@@ -1,9 +1,7 @@
 package enumeratum
 
-import scala.language.experimental.macros
-
 import scala.collection.immutable._
-import scala.collection.breakOut
+import scala.language.experimental.macros
 
 /**
   * All the cool kids have their own Enumeration implementation, most of which try to
@@ -40,19 +38,19 @@ trait Enum[A <: EnumEntry] {
     * Map of [[A]] object names to [[A]]s
     */
   lazy val namesToValuesMap: Map[String, A] =
-    values.map(v => v.entryName -> v)(breakOut)
+    values.map(v => v.entryName -> v).toMap
 
   /**
     * Map of [[A]] object names in lower case to [[A]]s for case-insensitive comparison
     */
   lazy final val lowerCaseNamesToValuesMap: Map[String, A] =
-    values.map(v => v.entryName.toLowerCase -> v)(breakOut)
+    values.map(v => v.entryName.toLowerCase -> v).toMap
 
   /**
     * Map of [[A]] object names in upper case to [[A]]s for case-insensitive comparison
     */
   lazy final val upperCaseNameValuesToMap: Map[String, A] =
-    values.map(v => v.entryName.toUpperCase -> v)(breakOut)
+    values.map(v => v.entryName.toUpperCase -> v).toMap
 
   /**
     * Map of [[A]] to their index in the values sequence.
--- a/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
@@ -31,8 +31,8 @@ object EnumEntry {
    *
    * http://stackoverflow.com/a/19832063/1814775
    */
-  private val regexp1: Pattern = Pattern.compile("([A-Z]+)([A-Z][a-z])")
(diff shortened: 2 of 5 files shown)


enumeratum-core/src/test/scala/enumeratum/Models.scala
```diff
@@ -227,6 +227,19 @@ object UncapitalisedEnum extends Enum[UncapitalisedEnum] {
 
 }
 
+case object MultiEnum extends Enum[MultiEnum] {
+  import scala.collection.breakOut
+  val values = findValues
```
lloydmeta: You can't do this because the type parameter passed to `Enum` must be a sealed trait.

Compilation error https://travis-ci.org/lloydmeta/enumeratum/jobs/343456716#L844

```scala
[error] /home/travis/build/lloydmeta/enumeratum/enumeratum-core/src/test/scala/enumeratum/Models.scala:232:16: You can only use findValues on sealed traits or classes
[error]   val values = findValues
[error]                ^
```

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
```diff
@@ -167,6 +175,9 @@ object EnumEntry {
     */
   trait LowerCamelcase extends EnumEntry with Camelcase with Uncapitalised
 
+  abstract class MultiEnum(override val entryName: String,
```
lloydmeta: I don't think this is needed (at least not here); looks like something you want in your tests.

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
```diff
@@ -31,8 +31,8 @@ object EnumEntry {
    *
    * http://stackoverflow.com/a/19832063/1814775
    */
-  private val regexp1: Pattern    = Pattern.compile("([A-Z]+)([A-Z][a-z])")
-  private val regexp2: Pattern    = Pattern.compile("([a-z\\d])([A-Z])")
+  private val regexp1: Pattern = Pattern.compile("([A-Z]+)([A-Z][a-z])")
```
lloydmeta: Since we don't need the class you added, please revert these formatting changes.

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -38,20 +39,20 @@ trait Enum[A <: EnumEntry] {
   /**
     * Map of [[A]] object names to [[A]]s
     */
-  lazy final val namesToValuesMap: Map[String, A] =
-    values.map(v => v.entryName -> v).toMap
+  lazy val namesToValuesMap: Map[String, A] =
+    values.map(v => v.entryName -> v)(breakOut)
```
lloydmeta: I'm not a fan of using `breakOut` here; I know what it does, but:

  1. It's not obvious to beginners how it works; 
  2. It'll likely be gone with the [new collections lib](https://www.scala-lang.org/blog/2017/02/28/collections-rework.html)
  3. This is a cached value; the so gains are pretty minimal.

enumeratum-core/src/test/scala/enumeratum/Models.scala
neilchaudhuri: Fixed.

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
neilchaudhuri: Fixed.

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
neilchaudhuri: Fixed.

enumeratum-core/src/main/scala/enumeratum/Enum.scala
neilchaudhuri: tl;dr Removed `breakOut`.

`breakOut` is definitely complex since `CanBuildFrom` isn't exactly the most intuitive concept, and I know that its days are numbered in Scala collections--though it isn't deprecated and probably won't be for some time. It just won't be necessary with the new views. 

Also, `breakOut` is a memory optimization to eliminate the need for an intermediate sequence--cached values or not. I worry that might matter when people use enums in "Big Data" (as with Apache Spark) scenarios, but it is definitely true that the sequences for most cases are so small that `breakOut` probably doesn't change the world.

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -38,7 +37,7 @@ trait Enum[A <: EnumEntry] {
   /**
     * Map of [[A]] object names to [[A]]s
     */
-  lazy final val namesToValuesMap: Map[String, A] =
+  lazy val namesToValuesMap: Map[String, A] =
```
lloydmeta: Nice. 

I think we should probably modify these `Map`s to be in terms of `namesToValuesMap` as well (I'd like to keep them final for now)

https://github.com/lloydmeta/enumeratum/blob/c895c327ba48e7646cf8c5815b505efd7d107fd6/enumeratum-core/src/main/scala/enumeratum/Enum.scala#L47-L54

lloydmeta:
Actually, the tests are failing with valid failures, 

See: https://travis-ci.org/lloydmeta/enumeratum/jobs/343869470#L2751

```

[info]   - should return the proper object when passed the proper string *** FAILED ***
[info]     java.util.NoSuchElementException: One is not a member of Enum (one, two)
[info]     at $c_ju_NoSuchElementException.$c_jl_Throwable.fillInStackTrace__jl_Throwable(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:27434:14)
[info]     at $c_ju_NoSuchElementException.$c_jl_Throwable.init___T__jl_Throwable(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:27466:8)
[info]     at java.util.NoSuchElementException.<init>(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:43347:52)
[info]     at enumeratum.Enum.withName(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:1193:46)
[info]     at {anonymous}()(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:75695:27)
[info]     at scala.scalajs.runtime.AnonFunction0.apply(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:36528:23)
[info]     at org.scalatest.OutcomeOf.outcomeOf(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:1852:7)
[info]     at org.scalatest.Transformer.apply(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:48096:10)
[info]     at org.scalatest.FunSpecLike$$anon$1.apply(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:38048:72)
[info]     at org.scalatest.FunSpecLike.invokeWithFixture$1(/home/travis/build/lloydmeta/enumeratum/enumeratum-core/.js/target/scala-2.12/enumeratum-test-fastopt.js:71598:15)
```

Pretty sure that's a fresh failure. Have you tried running the tests locally??

neilchaudhuri:
I ran `test` and got this: "Passed: Total 63, Failed 0, Errors 0, Passed 63." I must be using sbt wrong.

I will figure out what's up with the failing test and get back to you.

enumeratum-core/src/main/scala/enumeratum/Enum.scala
neilchaudhuri: Can you clarify what you mean here?

neilchaudhuri:
Everything should be in order except with regard to that one clarification I need. Can you elaborate on that? Thanks.

lloydmeta:
@neilchaudhuri sorry for the radio silence I'm currently away on a business trip. Will take a closer look and answer any questions when I'm back !!

neilchaudhuri:
Oh OK. Good luck with the business and safe travels!

enumeratum-core/src/main/scala/enumeratum/Enum.scala
lloydmeta: Right now, they're built in terms of `values` directly, whereas I think they should be built in terms of `namesToValuesMap` so that they can have the same values (in case the user overrides `namesToValuesMap` and wants to use functions that call `lowerCaseNamesToValuesMap`)

neilchaudhuri: I think I got it.

lloydmeta:
I've just released this as part of 1.5.13 :)

neilchaudhuri:
Awesome. Thanks. I hope people find the feature useful.

drewboardman:
Is there any documentation on how to use this? I'm not seeing `MultiEnum` available. Was this removed - or are we intended to also override `nameToValuesMap`?

lloydmeta:
https://github.com/lloydmeta/enumeratum/blob/87c4096ef02df15539ff7a25baa26fce25bb5f40/enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala#L438-L442

^ should show usage?

ChernikovP:
@drewboardman , apparently this `MultiEnum` is still and was always under `test` directory, rather than `src`, thus it's not actually available...

PR #176, opened 2018-02-20
Add support of jsoniter-scala codecs

plokhotnyuk:
For users of enumeratum and jsoniter-scala (https://github.com/plokhotnyuk/jsoniter-scala)

### DIFF
--- a/enumeratum-jsoniter-scala/src/main/scala/enumeratum/values/JsoniterScala.scala
+++ b/enumeratum-jsoniter-scala/src/main/scala/enumeratum/values/JsoniterScala.scala
@@ -42,8 +42,15 @@ object JsoniterScala {
           enum.withValueOpt(v).getOrElse(in.enumValueError(v.toString))
         }
 
+      def decodeKey(in: JsonReader): A = {
+        val v = in.readKeyAsInt()
+        enum.withValueOpt(v).getOrElse(in.enumValueError(v.toString))
+      }
+
       def encode(x: A, out: JsonWriter): Unit =
         if (x ne null) out.writeVal(x.value.toInt) else out.writeNull()
+
+      def encodeKey(x: A, out: JsonWriter): Unit = out.writeKey(x.value.toInt)
     }
 
   def codec[A <: LongEnumEntry: Manifest](enum: LongEnum[A]): JsonCodec[A] =
@@ -60,8 +67,15 @@ object JsoniterScala {
           enum.withValueOpt(v).getOrElse(in.enumValueError(v.toString))
         }
 
+      def decodeKey(in: JsonReader): A = {
+        val v = in.readKeyAsLong()
+        enum.withValueOpt(v).getOrElse(in.enumValueError(v.toString))
+      }
+
       def encode(x: A, out: JsonWriter): Unit =
         if (x ne null) out.writeVal(x.value.toLong) else out.writeNull()
+
+      def encodeKey(x: A, out: JsonWriter): Unit = out.writeKey(x.value.toLong)
     }
 
   def codec[A <: ShortEnumEntry: Manifest](enum: ShortEnum[A]): JsonCodec[A] =
@@ -78,8 +92,15 @@ object JsoniterScala {
           enum.withValueOpt(v).getOrElse(in.enumValueError(v.toString))
         }
 
+      def decodeKey(in: JsonReader): A = {
+        val v = in.readKeyAsShort()
+        enum.withValueOpt(v).getOrElse(in.enumValueError(v.toString))
+      }
+
       def encode(x: A, out: JsonWriter): Unit =
         if (x ne null) out.writeVal(x.value.toShort) else out.writeNull()
+
+      def encodeKey(x: A, out: JsonWriter): Unit = out.writeKey(x.value.toShort)
     }
 
   def codec[A <: StringEnumEntry: Manifest](enum: StringEnum[A]): JsonCodec[A] =
@@ -96,8 +117,15 @@ object JsoniterScala {
           enum.withValueOpt(v).getOrElse(in.enumValueError(v))
         }
 
+      def decodeKey(in: JsonReader): A = {
+        val v = in.readKeyAsString()
+        enum.withValueOpt(v).getOrElse(in.enumValueError(v))
+      }
+
       def encode(x: A, out: JsonWriter): Unit =
         if (x ne null) out.writeVal(x.value.toString) else out.writeNull()
+
+      def encodeKey(x: A, out: JsonWriter): Unit = out.writeKey(x.value.toString)
     }
 
   def codec[A <: ByteEnumEntry: Manifest](enum: ByteEnum[A]): JsonCodec[A] =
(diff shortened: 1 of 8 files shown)


plokhotnyuk:
Got test failure for 2.12.x build, should it be fixed in the ScalaJS tests of Circe integration or just need to restart the build?

Here is as it was reported in the build log:
`Could not run test enumeratum.CirceSpec: java.util.concurrent.TimeoutException: Futures timed out after [1 minute]`

enumeratum-jsoniter-scala/src/main/scala/enumeratum/values/JsoniterScala.scala
```diff
@@ -0,0 +1,138 @@
...
+    new JsonCodec[A] {
+      def nullValue: A = null.asInstanceOf[A]
+
+      def decode(in: JsonReader, default: A): A =
+        if (in.isNextToken('n')) {
+          in.rollbackToken()
+          in.readNullOrError(default, "expected number value or null")
+        } else {
+          in.rollbackToken()
+          val v = in.readInt()
```
lloydmeta: It **looks** like these codecs are all the same except for 

```
val v = in.read$Type

// and 

out.writeVal(x.value.to$Type)
```

If so, maybe we can make this code more DRY?

plokhotnyuk: It is possible to generalize and extract a common part of implementation but it will lead to code that more inefficient in runtime.
Better option would be generation of specialized enums by Scala macros, will it be acceptable from your side?

lloydmeta: Aah, ok. I took at look at `jsoniter` and I see that you're aiming at high performance.

Just a thought though, why not just publish this lib yourself?

plokhotnyuk: It happened that from version 0.30.x the `make` macro of `jsoniter-scala` generates codecs for  enumeratum enums that are represented as case objects with sealed base in Scala and as strings in JSON.

These codecs have performance characteristics that are much greater then when using enumeratum methods in run-time.

Here are results of benchmarks which compares parsing and serialization of enumeratum enums for Circe, Play-JSON and jsoniter-scala libraries on my notebook with JDK 11 RC:

```
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                                                  (size)   Mode  Cnt        Score       Error  Units
[info] ArrayOfEnumeratumADTsBenchmark.readCirce                      128  thrpt    5    70160.024 ±  2600.274  ops/s
[info] ArrayOfEnumeratumADTsBenchmark.readJsoniterScala              128  thrpt    5   386560.693 ± 24150.781  ops/s
[info] ArrayOfEnumeratumADTsBenchmark.readPlayJson                   128  thrpt    5    71450.235 ±  4748.745  ops/s
[info] ArrayOfEnumeratumADTsBenchmark.writeCirce                     128  thrpt    5    31986.244 ±   323.127  ops/s
[info] ArrayOfEnumeratumADTsBenchmark.writeJsoniterScala             128  thrpt    5  1964112.081 ± 61708.733  ops/s
[info] ArrayOfEnumeratumADTsBenchmark.writeJsoniterScalaPrealloc     128  thrpt    5  2125000.430 ± 10822.709  ops/s
[info] ArrayOfEnumeratumADTsBenchmark.writePlayJson                  128  thrpt    5   212672.311 ±  6918.046  ops/s
```

As you can see it serializes with speed ~2 CPU cycles per byte in JSON representation (or ~2Gb per second on my notebook).

@lloydmeta Please review sources of these benchmarks: 
https://github.com/plokhotnyuk/jsoniter-scala/pull/152/files

lloydmeta: Interesting benchmarks! It looks promising enough that it pretty much makes this PR unnecessary :)

A couple thoughts on my side are:

1. Why are you specifically benchmarking deserialising an array of all Enum member? In my experience, by far the most common use-case is to deserialise a single enum member as part of a bigger JSON tree?
2. The Circe read benchmark is `def readCirce(): Array[EnumeratumSuit] = decode[Array[EnumeratumSuit]](new String(jsonBytes, UTF_8)).fold(throw _, x => x)`, but ... you already have `jsonString`? It seems to me this might explain the lower throughput when trying to compare performance between Circe and Jsoniter.
3. Similarly, for the Circe write benchmark, you're making it use a pretty printer?
4. For that matter .. it seems to me like all of these are purely focused on reading from and writing to bytes...but I'm not sure that's what Circe is optimised for?

Then there are the various optimisations you use in Jsoniter...like a thread local writer and specialised types for generic methods. I'm not sure what you mean by "review", but after taking a brief look, I'm not sure what conclusions to draw ?

plokhotnyuk: @lloydmeta Lot of thanks for your reviews!

Yeh, changes that are proposed in this PR are unnecessary now.

1. You are right that parsing and serialization of values should be tested in isolation as much as possible, but parsing of single values or dozen values of case class fields will show worse results (for per value rates). Please see results of `ArrayOfIntsBenchmark` and `IntBenchmark` here: 
http://jmh.morethan.io/?source=https://plokhotnyuk.github.io/jsoniter-scala/oraclejdk11.json

2. Usually, we parse JSON from buffered UTF-8 bytes (not Java strings) that are received from the wire or disks, so `jsoniter-scala` benchmarks are more representative and take in account a whole pipeline of parsing and serialization of data. Moreover `circe` has an API for parsing immediately from bytes but it is slower than the manual decoding of `String` instances and using an API for them.

3.  `pretty` is a misleading method name that is used for both: no space printing and pretty printing. But noSpace option (or compact, no space serialization) is used for all (not only for `circe`) benchmarks. Also, it is checked by tests for benchmarks.

4. Yes, `jsoniter-scala` parses and serializes only from/to bytes or Input/Output streams, but not only for greater performance... it is also to avoid encoding problems, when standard Java/Scala decoders replace malformed UTF-8 bytes by placeholders or do not flag broken surrogate pairs of UTF-16 characters.

Thanks again for your feedback!

And welcome to [our chat](https://gitter.im/plokhotnyuk/jsoniter-scala) or [our issue board](https://github.com/plokhotnyuk/jsoniter-scala/issues) - any your input will be greatly appreciated!

lloydmeta: > BTW, thread locals are not hacks - even `circe` use them for a competitive advantage (see `reuseWriters` option of printers).

Right..no one mentioned anything about "hacks". The point I was trying to allude to is that the results from the benchmarks you posted aren't testing anything in isolation and therefore it's hard to draw conclusions.  

At the end of the day, the only thing they conclusively show is jsoniter-scala being really fast at the use-case you've focused on, which is great and speaks volumes about the optimisations that you've put into the project.

PR #178, opened 2018-02-26
Circe KeyEncoder/KeyDecoder for String Enum

alonsodomin:
Thought that it was a nice addition to the `CirceStringEnum` to be able to be used as a key in `Map`s (i.e.: `Map[MyEnum, Int]`).

For that, Circe requires a `KeyEncoder` and a `KeyDecoder` to be present in the implicit scope. With current enumeratum's API, implementing those is trivial for enums that are based on strings, so it could be cool to provide it out of the box.


lloydmeta:
Just released this in `1.5.16` of enumeratum-circe. Thanks again !

PR #181, opened 2018-03-19
Fix scalacheck dependencies

nrinaudo:
I'm trying to think of something witty, but... yeah. Just look at the diff. My bad. Sorry.


lloydmeta:
That's amazing ! I guess it worked because the version matched perfectly btwn Circe and Scalacheck ??

Sorry I didn't catch it during review 😭

lloydmeta:
Just released a version with your fix ! Thanks again!

PR #184, opened 2018-03-22
added support for case insensitive PlayJson de/serialisation

dbuschman7:


enumeratum-play-json/src/test/scala/enumeratum/Dummy.scala
```diff
@@ -10,3 +10,16 @@ object Dummy extends Enum[Dummy] with PlayJsonEnum[Dummy] {
   case object c extends Dummy
   val values = findValues
 }
+
+/**
+  * Created by dbuschman on 03/20/2018
+  */
+sealed trait InsensitiveDummy extends EnumEntry
```
lloydmeta: This name is genius. Insensitive Dummy 😆

lloydmeta:
Thanks; done and released as 1.5.14 of `enumeratum-play` and `enumeratum-play-json`

enumeratum-play-json/src/test/scala/enumeratum/Dummy.scala
dbuschman7: Yep! It was DummyInsensitive for about 10 seconds, but then ...

Issue #135, opened 2017-04-18
Potential addition to Slick integration for easy MappedColumnType[ValueEnum] definitions


kolemannix:
First off, I want to say thank you for Enumeratum, it really is The Missing Scala Enum Library (TM), and I wish I had discovered it sooner.

I just thought I would share a little helper trait and function I've been using for defining `MappedColumnTypes` for instances of `ValueEnum[A]` with less boilerplate:

```scala
import scala.reflect.ClassTag
import enumeratum.values.{ ValueEnum, ValueEnumEntry }
import slick.driver.JdbcProfile

trait Profile {
  val profile: JdbcProfile
}

trait MappedColumnTypeEnumSupport { this: Profile =>
  def mappedColumnTypeForValueEnum[V, E <: ValueEnumEntry[V]]
  (companion: ValueEnum[V, E])
  (implicit tag: ClassTag[E], v: profile.BaseColumnType[V]): profile.BaseColumnType[E] = {
    profile.MappedColumnType.base[E, V](
      { x => x.value },
      { v => companion.withValue(v) }
    )
  }
}
```
Usage (assuming a `ValueEnum` called `Shape` somewhere in scope): 
```scala
trait MyMappers extends MappedColumnTypeEnumSupport {
  implicit lazy val shapeMapping: BaseColumnType[Shape] = mappedColumnTypeForValueEnum(Shape)
}
```

Seems like this is generic enough that it could potentially be included with the library to ease this use case?If so, I believe something similar could be done for the `SetParameter` definitions as well.

Thanks again for this great library!

lloydmeta:
Hi there,

Thanks for your kind feedback :)

Your Slick integration looks good to me; I think it would be great to have that included so users don't need to write that boilerplate themselves. 

Would you be interested in adding such an integration?

kolemannix:
@lloydmeta 
I'd be happy to add an integration. I'll mention though, that I am a fan of separating Serialization concerns such as a `BaseTypedType` from definition of the data model. Thus, if I wrote an integration, it would probably not be of the mix-in style where the trait is mixed right in at the site the enum is defined. It just seems to couple them very tightly, and does not work if you have no control of the call-site, say if I wanted a `BaseTypedType` for an enum defined in a 3rd party library.

So more of a typeclass-instance-style interaction, where instances are defined in an ad-hoc fashion separately.

[...]

PR #187, opened 2018-06-08
(Issue #135): Slick support

kolemannix:
Finally got around to an attempt at #135.

Still todo:

- [x] Fix build for Scala 2.10.x (Slick 3.2 is not built for 2.10.x)
- [x] Clean up tests
- [x] Scaladoc
- [x] README section
- [x] Provide constructors for `Option[SetParameter[_]]` instances
- [x] Provide constructors for `GetResult[_]` and `Option[GetResult[_]]` instances

Looking for feedback on the general approach. Obviously its a bit different from the other integrations thanks to Slick requiring a `profile` in scope for defining column types.


kolemannix:
> It's a bit of a shame that Slick requires the profile to build type mappers (I like how quill has the option to create one without).

Completely agreed. I also went down the road of implementing a singleton first, then implementing the mixin-style in terms of the singleton, but as you've pointed out, it ends up being clunkier to use than simply defining the mappers directly.

I don't love the idea of requiring implementing code to have a trait with a member named `profile`, which is why providing a singleton as well would be nice. But having some form of `trait Profile { val profile: Profile }` is a pretty ubiquitous pattern in Slick so I think it's reasonable.

kolemannix:
@lloydmeta Any ideas on the cause of the build failure for only Scala 2.11? 

`Scala instance doesn't exist or is invalid`

https://travis-ci.org/lloydmeta/enumeratum/jobs/393739949#L606

lloydmeta:
@kolemannix Wow, never seen that error before 😨  

Googling around seems to suggest that this is a problem caused by discrepancies between the Scala version and sbt's Scala version (no idea why); can you try bumping one or the other?

lloydmeta:
This is looking like a weird corrupt jar / cache problem on Travis.  I cleared the build caches and now 2.11.12 is passing, but 2.10.7 fails 😓 

https://travis-ci.org/lloydmeta/enumeratum/builds/394125037

kolemannix:
Looks like it is all passing; feel free to review it now.

There is a singleton-based API for the `SetParameter [_]` and `GetResult[_]` stuff since those do not require a profile (they assume JDBC).

Because of this, I split column support away from plain sql support (i.e., `SlickEnumPlainSqlSupport` vs `SlickEnumColumnSupport`), but there is still an option to combine both in one shot by simply using `SlickEnumSupport` or `SlickValueEnumSupport`.

.travis.yml
```diff
@@ -19,7 +19,7 @@ matrix:
         - sbt -jvm-opts .travis-jvmopts ++$TRAVIS_SCALA_VERSION scalafmt::test test:compile test:doc
         - travis_wait 30 sleep 1800 &
         - sbt -jvm-opts .travis-jvmopts ++$TRAVIS_SCALA_VERSION test
-    - scala: 2.12.4
+    - scala: 2.12.6
```
kolemannix: Is this at all worrisome? It's probably not necessary to keep; I think the travis cache was the issue. But if it's beneficial to bump the 2.12 version here I will of course leave it in.

lloydmeta: No not at all. Please leave it in. 

It’s a bit late here so I’ll do a more thorough review before merging tmw but so far looking great :)

README.md
```diff
@@ -937,75 +939,66 @@ ctx.run(query[Shirt]).foreach(println)
...
-For example when you want the `Enum[Greeting]` defined in the introduction as a database column, you can use the following code
-
+In order to use your enumeratum Enums in Slick tables as columns, you will
+ need to construct instances of `MappedColumnType` and make them available
+ where you define and query your slick tables. In order to more easily
+ construct these instances, the enumeratum-slick integration provides a trait
+ `enumeratum.SlickEnumSupport`. This trait provides a method `mappedColumnTypeForEnum`
+ (and variants) for constructing a mapped column type for your enum. For example
+ if you want to use `Enum[Greeting]` in your slick table, mix in `SlickEnumSupport`,
+ which requires a slick `profile` to be in scope, and construct a mapper
```
lloydmeta: Can you add a small note here to make it obvious that this trait is typically mixed into where the `Table`s reside?

If I'm getting this right, it might be something like " which requires a slick `profile` to be in scope (typically where your `Table`s are defined).

I'm just hoping to make it obvious that the usage of this trait is different from that of the other integrations, where you would typically mix the support trait into the `Enum[T]`.

kolemannix: > If I'm getting this right, it might be something like " which requires a slick profile to be in scope (typically where your Tables are defined).

Sounds good. I'll add a fuller example in that style (with the tables) as well as one in the style of a separate mappers trait `trait CommonColumnMappers { val profile: Profile; ... }`. I think that should cover 99% of how people organize their slick code.

lloydmeta:
Thanks @kolemannix for the great work ! Just merged and pushed a release (1.5.15) to Maven Central.

PR #195, opened 2018-09-27
Add cats module

ybasket:
This PR adds support for some basic type classes from `cats-core`:
- `Eq` for typesafe comparison
- `Show` for pretty printing
- `Hash` for abstracting of hashing of values

Corresponding enumeratum-style mixins are provided for these, supporting both `Enum` and `ValueEnum`. The second also comes in an extended version which provides an instance of `Order`.

Major use cases are generic derivation and operations on collections of (nested) enum values, for example sorting a list of errors by their (enum) priority. See `Readme.md` for small examples.

### DIFF
--- a/enumeratum-cats/src/main/scala/enumeratum/CatsEnum.scala
+++ b/enumeratum-cats/src/main/scala/enumeratum/CatsEnum.scala
@@ -1,0 +1,14 @@
+package enumeratum
+
+import cats.instances.string._
+import cats.{Eq, Hash, Show}
+
+trait CatsEnum[A <: EnumEntry] { this: Enum[A] =>
+
+  implicit val eqInstance: Eq[A] = Eq.fromUniversalEquals[A]
+
+  implicit val showInstance: Show[A] = Show.fromToString[A]
+
+  implicit val hashInstance: Hash[A] = Hash.by(_.entryName)
+
+}
--- a/enumeratum-cats/src/main/scala/enumeratum/CatsEnum.scala
+++ b/enumeratum-cats/src/main/scala/enumeratum/CatsEnum.scala
@@ -1,14 +1,22 @@
 package enumeratum
 
-import cats.instances.string._
 import cats.{Eq, Hash, Show}
 
 trait CatsEnum[A <: EnumEntry] { this: Enum[A] =>
 
-  implicit val eqInstance: Eq[A] = Eq.fromUniversalEquals[A]
+  /**
+    * `Eq` instance for the enum entries - treats all enum values as distinct.
+    */
+  implicit val eqInstance: Eq[A] = Cats.eqForEnum[A]
 
-  implicit val showInstance: Show[A] = Show.show(_.entryName)
+  /**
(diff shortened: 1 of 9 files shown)


lloydmeta:
Nice! 

Do you think we should include a ScalaJS version of this too?

ybasket:
Appearently, the Slick tests are a bit flaky on Travis (it only failed on Scala 2.11 and I don't touch Slick at all) - ran the full test suite locally, all green. That seems also to be the reason the coverage report looks so bad.

enumeratum-cats/src/main/scala/enumeratum/CatsEnum.scala
```diff
@@ -0,0 +1,14 @@
+package enumeratum
+
+import cats.instances.string._
+import cats.{Eq, Hash, Show}
+
+trait CatsEnum[A <: EnumEntry] { this: Enum[A] =>
+
+  implicit val eqInstance: Eq[A] = Eq.fromUniversalEquals[A]
+
+  implicit val showInstance: Show[A] = Show.fromToString[A]
```
lloydmeta: Is this what we want? (e.g. rather than `_.entryName`)

enumeratum-cats/src/main/scala/enumeratum/values/CatsValueEnum.scala
```diff
@@ -0,0 +1,28 @@
+package enumeratum.values
+
+import cats.{Eq, Order, Show}
+
+trait CatsValueEnum[ValueType, EntryType <: ValueEnumEntry[ValueType]] {
+  this: ValueEnum[ValueType, EntryType] =>
+
+  implicit val eqInstance: Eq[EntryType] = Eq.fromUniversalEquals[EntryType]
```
lloydmeta: Would it make sense to have these depend on `Eq` on the `ValueType` and build an `Eq` for the `EntryType` using that?

ybasket: Interesting question - I tend to say no. The idea of an enum is having distinct values and these are guaranteed (correct me if I'm wrong) by the way the macros work, requiring literals in the definition. So the only valid `Eq` instance would be the one differentiating all enum values and we achieve that already by `fromUniversalEquals`. One can still have a custom instance by overriding this one, it's a one-liner.

So I'd leave it as-is for now, but arguments welcome :)

ybasket:
@lloydmeta Maybe we can add a new section in `Readme.md` on how to implement a new module stating such best practices? I should have seen this on Circe, but I guess I'm not the only failing there ;)

Anyway, thanks for the helpful and quick feedback! I updated the PR to provide this and also added some Scaladoc and a Readme section on how to use the Cats module without inheritance. These objects also provide methods to create instances of (for example) `Eq` based on a custom `Eq`, so now one can now more easily decide to use the default or something different.

lloydmeta:
Released as 1.5.14; should be up on Maven Central soon :) Thanks again !

PR #196, opened 2018-09-27
Upgrade circe

alonsodomin:


lloydmeta:
> mmm, we can't build against circe 0.10.x in scala 2.10...

Yeah, we'll need something like this https://github.com/lloydmeta/enumeratum/blob/390963c7b5cb183f23f417a0a92c7cb3cf6a4047/build.sbt#L18-L24

lloydmeta:
Thanks for this, just released this as v1.5.18 of enumeratum-circe.

PR #215, opened 2019-02-01
bump to play 2.7 that supports scala 2.13

jtjeferreira:


OlegYch:
this seems to use play 2.7 only on scala 2.13? how about adding 2.12?

jtjeferreira:
> this seems to use play 2.7 only on scala 2.13? how about adding 2.12?

You can always use play 2.7 when in scala 2.12 by yourself. I think the situation is:
* scala **2.12**: minimum play version **2.6**
* scala **2.13**: minimum play version **2.7**

OlegYch:
pretty sure they are not binary compatible and enumeratum has to be compiled with 2.7 if you want to use it with 2.7

OlegYch:
as it stands it would be impossible to use enumeratum-play(json) with play 2.7 and scala 2.12

lloydmeta:
Ah I see. I haven’t checked but I assume Play 2.7 also supports 2.12?

OlegYch:
@lloydmeta yep, and 2.11

lloydmeta:
I see. Would you like to send a PR to addresss the oversight?

OlegYch:
@lloydmeta will try, i suggest dropping support for play <2.7 and scala 2.10 (since play 2.7 doesn't support it)

lloydmeta:
> @lloydmeta will try, i suggest dropping support for play <2.7 and scala 2.10 (since play 2.7 doesn't support it)

It might be nice if we could keep 2.10 (just not bump its Play version) around if there aren't any significant code-level changes needed, but I guess we won't know for sure until we try :)

PR #217, opened 2019-02-06
use play 2.7 for all scalas > 2.10

OlegYch:
note this drops support for play 2.6


lloydmeta:
Done, just released these so they should be on Maven Central in a few hours :)

PR #219, opened 2019-02-18
Circe KeyEncoder/KeyDecoder for Enum

yaroot:
It will be nice to have key encoder/decoder for non-value based enums, since there is one for string enum.


yaroot:
@lloydmeta no it's done

lloydmeta:
Awesome. Thanks @yaroot I just published this as `enumeratum-circe` 1.5.21. Should be on Maven Central before long.

Issue #151, opened 2017-09-11
LowerCamelcase doesn't work


1gnition:
I'm using version `"com.beachape" %% "enumeratum" % "1.5.12"` with Scala 2.12, and LowerCamelcase prints the names with everything lowercase, no capitals at all.

Given the following example, I would expect `howAreYou` to be printed, but what's printed is `howareyou`:

```scala
import enumeratum.{Enum, EnumEntry}
import enumeratum.EnumEntry.LowerCamelcase

import scala.collection.immutable.IndexedSeq

sealed trait Thing extends EnumEntry with LowerCamelcase
object Thing extends Enum[Thing] {
  val values: IndexedSeq[Thing] = findValues
  case object HowAreYou extends Thing

}

println(Thing.HowAreYou.entryName)
```

lloydmeta:
Hi there,

Thanks for taking a look at enumeratum and filing this issue.

Things get a bit tricky with the different stackable modifiers here, but I believe this is how it was intended to be used:

https://github.com/lloydmeta/enumeratum/blob/e42b8d9d19b8aeff0c34d3f5066e5ec5e93235f1/enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala#L312

That trait was contributed by @dmit, so if you'd like to see it changed, maybe we should all discuss further :)

dmit:
Yep, `LowerCamelCase` was only meant to be used with `snake_case`-formatted enums, since it builds on `CamelCase`.

@1gnition, I think the `Uncapitalised` trait should do what you need in that example.

1gnition:
Thanks guys. `Uncapitalised` does the job

lukaszlenart:
I was hit by the same issue and using `Uncapitalised` did the job. Anyway I think you should change description of that class which is misleading now:

```
  /**
    * Stackable trait to convert the entryName to lowerCamelCase.
    */
```

It clear says that `entryName` will represented as a `camelCase` and there is no information that it works only with `*_*` pattern.

And the `LowerCamelcase` can be easily fixed:

```scala
trait LowerCaseValue extends EnumEntry {

  private val regexp1: Pattern    = Pattern.compile("([A-Z]+)([A-Z][a-z])")
  private val regexp2: Pattern    = Pattern.compile("([a-z\\d])([A-Z])")
  private val replacement: String = "$1_$2"

  private def camel2WordArray(name: String): Array[String] = {
[...]

PR #223, opened 2019-03-28
Changed logic of CamelCase and LowerCamelCase EnumEntries

ulanzetz:
I think is not right logic when entry FooBar have "Foobar" entryName in CamelCase enum and "foobar" entry in LowerCamelCase enum.
It can add support of transforming to CamelCase and LowerCamelCase not only delimited by "_" entries.

### DIFF
--- a/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
@@ -41,6 +41,20 @@ object EnumEntry {
     regexp2.matcher(first).replaceAll(replacement).split("_")
   }
 
+  private def word2Camel(str: String): String = {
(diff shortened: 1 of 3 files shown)


lloydmeta:
Interesting, I think this is basically what is talked about in https://github.com/lloydmeta/enumeratum/issues/151#issuecomment-328556146, the recommendation there seems to be to use `CamelCase` or `Lowercase`?

I'm open to "fixing" this, but the thing is, I don't want to have it so that every stackable trait works with every other one; the various possible combinations of inheritance make that a very difficult promise to uphold, and I want to make sure we don't break any existing expectations out there and/or introduce [more edge cases](https://github.com/lloydmeta/enumeratum/issues/151#issuecomment-470555399).

Issue #186, opened 2018-04-24
Worthwhile to Add Doobie Support?


kevinmeredith:
Would it be useful to include a `doobie` module within `enumeratum` that provides, `Meta[X]` where `X` is an `Enum` or `EnumEntry`?

I'm just about to write a `Meta` for an `enumeratum._` value, so that made me ask this question!

I'd be interested to work on this, but I wanted to post the question in case you've already implemented or thought about it.

Thanks!

lloydmeta:
Definitely sounds useful to me :) I think such a module might look quite similar to the Quill module, in case that helps.

sebastianvoss:
@kevinmeredith, yes this would be really cool :)

lloydmeta:
Welp, 10 thumbs ups probably means it should be done 😆

PR #227, opened 2019-05-02
Adds support for doobie

alexbalonperin:
closes https://github.com/lloydmeta/enumeratum/issues/186

The tests that I added only check whether there is a Read and Write for the enum types. They do not check however that the resulting values look good. The reason is to avoid adding extra dependencies like cats-effect and using an actual database like h2 or PostgreSQL. I could add them if necessary.

Also, for the version of the enumeratum-doobie module, I am not too sure where should it start.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,9 +8,10 @@ lazy val theScalaVersion = "2.12.7"
   2.13.0-M5 support is currently defined as a separate project (scala_2_13) for convenience while
   integration libraries are still gaining 2.13 support
  */
-lazy val scalaVersions     = Seq("2.10.7", "2.11.12", "2.12.7")
-lazy val scala_2_13Version = "2.13.0-M5"
-lazy val scalaVersionsAll  = scalaVersions :+ scala_2_13Version
+lazy val scalaVersions           = Seq("2.10.7", "2.11.12", "2.12.7")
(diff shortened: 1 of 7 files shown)


alexbalonperin:
Not sure why the `enumeratum-cats` tests failed. @lloydmeta Is this an intermittent failure or did I break something? 🤔

alexbalonperin:
I have tried cross-compiling the code to Scala 2.10. The latest available version of doobie supporting scala 2.10 seems to be 0.4.0. In 0.4.0, Meta was under `doobie.utils.meta`. I would have to write specific code for that version. 

Is there a way to avoid this? For example, by cross-compiling `enumeratum-doobie` only to scala 2.11 and above.

lloydmeta:
Nice! Thanks for the PR. I’ll take a closer look when I’m back from Jp holidays :)

The tests do have intermittent failures; they’re usually the 2.10 ones and are a known issue, unfortunately.

abalonperin:
@lloydmeta any update on this?

lloydmeta:
> @lloydmeta any update on this?

No, not at the moment; holidays are over but I'm travelling atm. Don't worry, I will come back to this.

lloydmeta:
Thanks for this! Just merged and released :)

Issue #216, opened 2019-02-06
publish additional JARs for 2.13.0-M5?


SethTisue:
followup to #199 

a lot more of your dependencies are now available, so it should be possible to publish e.g. enumeratum-circe, which we wanted over at https://github.com/pauljamescleary/scala-pet-store/pull/142

lloydmeta:
I just released the circe integration; with the new Play 2.7 integration that just got merged, I _think_ that's all for now?

Will keep this ticket open to track others.

Milyardo:
publishing for the release candidate now would be a great idea

lloydmeta:
> publishing for the release candidate now would be a great idea

Ah ok, looks like RC1 is a thing now https://github.com/scala/scala/milestones?state=closed

Unfortunately, we're a bit blocked at the moment on https://github.com/scalatest/scalatest/issues/1567 for the main core/macro projects. Very likely the same situation w/ the integration libs, but I'm more comfortable having breakage there in a separate branch for now as those projects start publishing artefacts for `2.13.0-RC1`.

Will keep an eye out though.

SethTisue:
> we're a bit blocked at the moment on scalatest/scalatest#1567

ScalaTest 3.0.8-RC2 has been published for Scala 2.13.0-RC1, is that sufficient for you to go forward with?

lloydmeta:
> > we're a bit blocked at the moment on [scalatest/scalatest#1567](https://github.com/scalatest/scalatest/issues/1567)
> 
> as per https://github.com/scala/make-release-notes/blob/2.13.x/projects-2.13.md, ScalaTest 3.0.8-RC2 has been published for Scala 2.13.0-RC1, is that sufficient for you to go forward with?

Yeah, I've now got a branch that tries to get us there https://github.com/lloydmeta/enumeratum/tree/build/2.13-rc1

I have a few more deprecation warnings and such to work through though.

guilgaly:
Hi, is there any news on this ? It looks like 2.13.0 final [might be here very soon](https://contributors.scala-lang.org/t/scala-2-13-0-rc2-release/3323/10).

lloydmeta:
> Hi, is there any news on this ? It looks like 2.13.0 final [might be here very soon](https://contributors.scala-lang.org/t/scala-2-13-0-rc2-release/3323/10).

[...]

PR #230, opened 2019-05-25
Scala 2.13.0-RC2 build

guilgaly:
See also issue #216.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
         - sbt -jvm-opts .travis-jvmopts ++$TRAVIS_SCALA_VERSION scalafmtCheck scalafmtSbtCheck test:compile test:doc
         - travis_wait 30 sleep 1800 &
         - sbt -jvm-opts .travis-jvmopts ++$TRAVIS_SCALA_VERSION test
-    - scala: 2.12.7
+    - scala: 2.12.8
       jdk: oraclejdk8
       script:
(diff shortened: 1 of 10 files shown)


lloydmeta:
Toggled this to see if we can't get Travis to build this..

lloydmeta:
Looks like there are some build failures. In the interest of putting out an RC-release, do you mind re-targeting this to https://github.com/lloydmeta/enumeratum/tree/build/2.13-rc1?

guilgaly:
I checked the Travis logs, apparently:

- Scala 2.10 build fails because the version of Wartremover doesn't exist for 2.10.
- Scala 2.11 build fails because of bad formatting in the `build.sbt` file (`scalafmtSbtCheck`); but I don't get why it's not an issue for the 2.12 build ?
- Scala 2.12 build succeeds.
- Scala 2.13 build fails because the various modules are missing the `enumeratum` and `enumeratum-test` dependencies. When testing locally I used `publishLocal` on the core modules and switched those dependency versions to the current snapshot version; I don't know how it's supposed to be handled normally ?

guilgaly:
Alright, fixed the `build.sbt` formatting issue, and now the Scala 2.11 build succeeds. I'll change the target to `build/2.13-rc1` now.

lloydmeta:
Thanks ! There are a few odd things that I think still need to be addressed (e.g. Play JSON doesn't have `2.8.0-M1` published for 2.11....) but I think this is enough to start allowing for RC builds on the branch.

PR #234, opened 2019-06-07
Scala 2.13.0

joan38:


lloydmeta:
Yep, looks like we'll need this to even begin :) https://github.com/scalatest/scalatest/issues/1601

This is the list I was talking about earlier.

https://github.com/lloydmeta/enumeratum/blob/b6d98ed95de4b3714494dd6d9184a303787be135/build.sbt#L100-L109

For now, it looks like Argonaut has `2.13` artefacts on Maven Central, but not the others ?

joan38:
I pushed with scalatest_2.13 3.0.8
It's complaining for enumeratum_2.13 itself not published in the tests

lloydmeta:
I think it's mostly a matter of making the versions in the build match with what each project needs, but I went ahead and pre-published the 2.13 arefacts for macro, core and test so the tests now pass.

All that needs to be done is to comment out the references here that don't have 2.13 artefacts:

https://github.com/lloydmeta/enumeratum/blob/b6d98ed95de4b3714494dd6d9184a303787be135/build.sbt#L100-L109

I think it should just be the Play related ones.

lloydmeta:
Thanks @joan38, just released all artefacts :)

joan38:
Thanks @lloydmeta

PR #241, opened 2019-07-02
Update circe version for publishing enumeratum-circe on 2.13 (jvm+js)

zakpatterson:
Hi, I noticed that enumeratum_circe has not published on 2.13 yet. Circe is available for 2.13 via `0.12.0-M4`. Tests pass on this version without modification.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -67,6 +67,7 @@ def thePlayJsonVersion(scalaVersion: String) =
 
 def theCirceVersion(scalaVersion: String) =
   CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, scalaMajor)) if scalaMajor >= 13 => "0.12.0-M4"
     case Some((2, scalaMajor)) if scalaMajor >= 11 => "0.11.1"
     case Some((2, scalaMajor)) if scalaMajor == 10 => "0.9.3"
     case _ =>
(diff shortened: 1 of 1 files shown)


zakpatterson:
This works for the jvm project, but not js. I'm not super experienced on the js side yet, so that'd be better left to someone that knows how to fix whatever issue is coming up there.

lloydmeta:
I'm following up with this on the Circe side to see why it fails with ScalaJS.

lloydmeta:
Travis Brown answered my question in https://github.com/circe/circe/pull/945#issuecomment-508724352

Basically 0.12 w/ ScalaJS requires you to bring your own `java.time._ implementation` (in [release notes](https://github.com/circe/circe/releases/tag/v0.12.0-M4))

So I **think** this just needs something like the following 

```scala
.jsSettings(
  libraryDependencies += "io.circe" %%% "not-java-time" % "0.2.0" % Test
)
```

added to the end of the `enumeratumCirce` module definition:

https://github.com/lloydmeta/enumeratum/blob/fabfd5ae3189653938410babb16c2e297e33685b/build.sbt#L316-L331

zakpatterson:
Thanks for the research @lloydmeta, I went ahead and added your suggestion, but filtered to only modify for 2.13.

zakpatterson:
I just retriggered the build via a force-push with no modifications. I'm not sure what caused the prior 2.10 testing error. Looks like this works.

lloydmeta:
Just released a 2.13 `enumeratum-circe` with this change :)

PR #246, opened 2019-07-26
Add proper support for postgres enum

yaroot:
Postgres enums are encoded with PgObject in jdbc (and doobie). PgObject is basically just a `(type: String, value: String)`pair, doobie-postgres has basic support for it. The approach will be different from other modules so I'd like to get some feedback.

Right now I have the very basic implementation, basically you need to feed a type name (in postgres) to `DoobiePgEnum.meta` and get back a `Meta` instance for your enum. I'd like to have a more convenient API but seems hard without using `lazy val` or macros.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -144,6 +144,7 @@ lazy val integrationProjectRefs = Seq(
   enumeratumQuillJs,
   enumeratumQuillJvm,
   enumeratumDoobie,
+  enumeratumDoobiePg,
   enumeratumSlick,
   enumeratumCatsJs,
   enumeratumCatsJvm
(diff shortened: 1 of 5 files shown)


lloydmeta:
Sorry it took a while for me to get back to you on this: I had to take a look at https://github.com/tpolecat/doobie/blob/f17d00fae2ba7a46d1dcb40b57a937a9a998abba/modules/postgres/src/main/scala/doobie/postgres/Instances.scala#L147-L165 to understand what was going on :)

I think this is a reasonable approach to supporting PG enums; can you add some tests, etc so we can try/hash out the API?

yaroot:
Unrelated tests are failing..

yaroot:
@lloydmeta sorry for disappearing, please have a look, basic support is there. Should this be in a separate package?

lloydmeta:
After giving it some thought, I think rather than this being something included in `enumeratum-doobie`, which adds support for `doobie-core`, I think this makes more sense as a separate module, given as how `doobie-postgres` is its own separate dependency.  Thoughts?

yaroot:
@lloydmeta done

lloydmeta:
@yaroot sorry I was unclear, what I meant was not for you to create another module in this project, but to publish it yourself.  I haven't myself used the doobie-postgres integration extensively enough to be a good steward for this integration.

yaroot:
alright, that's fine

PR #248, opened 2019-09-02
Cats to 2.0.0 and build enumeratum-cats  for scala 2.13

little-inferno:


lloydmeta:
Thanks for this @little-inferno . Just released 1.5.16 with this to Maven Central.

PR #249, opened 2019-09-10
bump reactivemongo version and make it publishable for scala 2.13

lambdista:


lloydmeta:
Thanks @lambdista :) Just pushed 1.5.14 to Maven Central

lambdista:
Thank you @lloydmeta! :)

PR #252, opened 2019-09-17
Remove dependency on circe not-java-time in JS enumeratum-circe

bastewart:
This is not needed as of circe 0.12.0

Confirmed in release notes: https://github.com/circe/circe/releases/tag/v0.12.0

circe/circe#1245 has discussion on it not being needed and resolution.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -336,13 +336,6 @@ lazy val enumeratumCirce = crossProject(JSPlatform, JVMPlatform)
       )
     }
   )
-  .jsSettings(libraryDependencies ++= (CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((2, scalaMajor)) if scalaMajor >= 13 =>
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -336,13 +336,6 @@ lazy val enumeratumCirce = crossProject(JSPlatform, JVMPlatform)
       )
     }
   )
-  .jsSettings(libraryDependencies ++= (CrossVersion.partialVersion(scalaVersion.value) match {
```
bastewart: I realised after committing that the `libraryDependencies` may still be needed? I was working off reverting the part of #241 that added `not-java-time`, but may have been too aggressive...

lloydmeta: I think this is fine; the only goal of this was to add `not-java-time` to the JS dependencies.

lloydmeta:
Thanks !

PR #258, opened 2019-10-31
Provide withName* variants returning an Either[String, EnumEntry]

pdalpra:
Similar to the Option variants, this provide safe  functions to get an EnumEntry by name, returning a Left is the entry cannot be found.
This reuses the helper function to builds the error message for Exception.

NB: this can eventually be reused in some integrations (Circe, Play) that uses Either in their APIs, which currently relies on the withName* Option variants and pattern matches on the result to convert it to Either.


pdalpra:
Indeed, I'll add right away !
Regarding the build failure, it seems that there is a transient issue on some Scala versions, but not always the same version :thinking:

pdalpra:
Is is okay for you ?

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -85,6 +85,13 @@ trait Enum[A <: EnumEntry] {
     */
   def withNameOption(name: String): Option[A] = namesToValuesMap.get(name)
 
+  /**
+    * Returns an [[Right[A]]] for a given name, or a [[Left[String]]] if the name does not match any of the values'
+    * .entryName values.
+    */
+  def withNameEither(name: String): Either[String, A] =
+    namesToValuesMap.get(name).toRight(buildNotFoundMessage(name))
```
lloydmeta: I've been thinking about this this some more, and I now think that instead of allocating a string for Left, it might be better (more typesafe, fewer allocations) if we returned an object (e.g. `case object NoSuchMember`).

pdalpra: It could be interesting, but it would mean that we would lose the error message that we currently get in Left.
WDYT of using a case class that could hold:

* a reference of the name we tried (that is already allocated in memory, since it was the parameter)
* a reference to the enumeration values (already in memory too)

`toString` could be overridden to produce the error message, only when needed.
There would still be allocations for the case class, but it should be cheaper than allocating the string.

lloydmeta: Indeed, that could work, and the cost of allocating a case class holding 2 references is cheap, so I'm 👍  to going in that direction.

Qthunder: What do you think of the case class extending Throwable. Could be of use to some users (`enableWritableStackTrace = false` if you're concerned with allocation)

Qthunder:
Was just about implement this very same feature :) Let me know if I can help

enumeratum-core/src/main/scala/enumeratum/Enum.scala
lloydmeta: I wouldn’t mind extending `Exception`, but only if there is no stack trace. We could for instance extend `scala.util.control.NoStackTrace`. 

My stance here is that if some one wants a stack trace, they can throw their own exception on left, it will not be part of this API.

pdalpra: Indeed, we could even go as far as extending `NoSuchElementException`, what do you think?

pdalpra:
I didn't get time recently to work on it, hopefully this week end :)

pdalpra:
Sorry for the delay, finally got time to update this PR, with the change discussed above:

* Introduce `NoSuchMember` case classes (one for Enum, one for ValueEnum) 
* Extend `NoSuchElementException` and `NoStackTrace`
* Override `getMessage` instead of using exceptions constructors taking a constructor to not allocate the string if not needed

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -85,6 +85,13 @@ trait Enum[A <: EnumEntry] {
     */
   def withNameOption(name: String): Option[A] = namesToValuesMap.get(name)
 
+  /**
+    * Returns an [[Right[A]]] for a given name, or a [[Left[String]]] if the name does not match any of the values'
```
lloydmeta:
```suggestion
    * Returns an [[Right[A]]] for a given name, or a [[Left[NoSuchMember]]] if the name does not match any of the values'
```

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -139,6 +146,27 @@ trait Enum[A <: EnumEntry] {
   def withNameLowercaseOnlyOption(name: String): Option[A] =
     lowerCaseNamesToValuesMap.get(name)
 
+  /**
+    * Returns an [[Right[A]]] for a given name, or a [[Left[String]]] if the name does not match any of the values'
+    * .entryName values, disregarding case.
+    */
+  def withNameInsensitiveEither(name: String): Either[String, A] =
```
lloydmeta: Can we get this updated to return `Left(NoSuchMember)` as well?

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -139,6 +146,27 @@ trait Enum[A <: EnumEntry] {
...
+    lowerCaseNamesToValuesMap.get(name.toLowerCase).toRight(buildNotFoundMessage(name))
+
+  /**
+    * Returns an [[Right[A]]] for a given name, or a [[Left[String]]] if the name does not match any of the values'
+    * .entryName values, disregarding case.
+    */
+  def withNameUppercaseOnlyEither(name: String): Either[String, A] =
```
lloydmeta: Same here.

enumeratum-core/src/main/scala/enumeratum/Enum.scala
```diff
@@ -139,6 +146,27 @@ trait Enum[A <: EnumEntry] {
...
+    upperCaseNameValuesToMap.get(name).toRight(buildNotFoundMessage(name))
+
+  /**
+    * Returns an [[Right[A]]] for a given name, or a [[Left[String]]] if the name does not match any of the values'
+    * .entryName values, disregarding case.
+    */
+  def withNameLowercaseOnlyEither(name: String): Either[String, A] =
```
lloydmeta: Same here

enumeratum-core/src/main/scala/enumeratum/values/ValueEnum.scala
```diff
@@ -62,6 +62,14 @@ sealed trait ValueEnum[ValueType, EntryType <: ValueEnumEntry[ValueType]] {
     */
   def withValueOpt(i: ValueType): Option[EntryType] = valuesToEntriesMap.get(i)
 
+  /**
+    * Returns an [[Right[EntryType]]] for a given value, or a [[Left[String]]] if the value does not match any of the values'
```
lloydmeta:
```suggestion
    * Returns an [[Right[EntryType]]] for a given value, or a [[Left[NoSuchMember]]] if the value does not match any of the values'
```

enumeratum-core/src/main/scala/enumeratum/Enum.scala
pdalpra: Indeed, completely missed those ones !

pdalpra:
All fixed, completely missed the documentation and case variants methods :sweat_smile:

lloydmeta:
Hey, thanks for this @pdalpra . Just released 1.5.15 based on this.

PR #263, opened 2019-12-11
Allow aliases in value enums

cb372:
Hello! 👋

I'd like to use enumeratum to model [protobuf enumerations](https://developers.google.com/protocol-buffers/docs/proto3#enum). Protobuf allows you to define aliases in your enums, so I've added a marker trait to allow me to disable the unique-values check in the macro.

WDYT?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -205,9 +205,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
     name := "enumeratum",
     version := Versions.Core.head,
     crossScalaVersions := scalaVersionsAll,
-    libraryDependencies += "com.beachape" %% "enumeratum-macros" % Versions.Macros.stable
+    //libraryDependencies += "com.beachape" %% "enumeratum-macros" % Versions.Macros.stable
   )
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -205,9 +205,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
     name := "enumeratum",
     version := Versions.Core.head,
     crossScalaVersions := scalaVersionsAll,
-    libraryDependencies += "com.beachape" %% "enumeratum-macros" % Versions.Macros.stable
+    //libraryDependencies += "com.beachape" %% "enumeratum-macros" % Versions.Macros.stable
   )
-//  .dependsOn(macros) // used for testing macros
+  .dependsOn(macros) // used for testing macros
```
cb372: Not sure what to do with this. If I put it back to how it was, the tests won't compile until a new version of the macros is published.

cb372:
The failing build looks like a flaky compiler bug. Just needs a rebuild.

lloydmeta:
Done. Released as `1.5.14`. Thanks !

PR #265, opened 2019-12-16
Adapting Enum Handlers to new reactive-mongo API

olsio:
reactive-mongo 0.19.x introduced a huge refactoring which is aimed to stabilize an API for version 1.0. The new API is not backward compatible and requires a rework of the previous enumeratum handlers.

This refactoring also removed support for Scala 2.10 for the new API libraries which is my 2.10 is also removed for this enumeratum module.


olsio:
I am not sure why the coverage dropped as the relevant module is covered 100% and the file with less coverage are outside of the changed module.

lloydmeta:
@olsio Thanks for this. I just released this as `1.5.15` of `enumeratum-reactive-mongo-bson`; it should be on Maven Central in a bit.

olsio:
Thanks

PR #266, opened 2019-12-25
Update doobie version to 0.8.8

jatcwang:


enumeratum-doobie/src/main/scala/enumeratum/Doobie.scala
```diff
@@ -1,6 +1,7 @@
 package enumeratum
 
 import doobie.util._
+import doobie.Meta
```
lloydmeta: Interesting ... a *patch* version bump resulted in breaking source compatibility??

jatcwang: I think it broke binary compatibility as well.. :/ My code wouldn't compile because enumeratum-doobie is on 0.8.6 while I upgraded to 0.8.8. (Complains about missing doobie.util.Meta in classpath)

lloydmeta: Yikes. Thanks.

lloydmeta:
@jatcwang thanks for this. Just released it as 1.5.17. It should be on Maven Central soon.

PR #269, opened 2020-01-11
Adds Play 2.8.x support, drops Scala 2.10 support

a1kemist:
In play-json `2.8.x`, the writes trait was [changed](https://github.com/playframework/play-json/pull/216) from `Writes[-T]` (contravariant) to `Writes[T]` (invariant) which breaks enumeratum's `PlayJsonEnum` trait. This PR does the following:

- Adds an implicit helper to contraMap the `Writes` provided by the existing trait. This seems to be sufficient enough to support play-json `2.8.x` without any more complicated code changes.
- Drops Scala 2.10 support. While working on this PR there were a few sporadic dependency resolution issues with resolving 2.10 based dependencies, but now I'm unable to resolve any of the 2.10 dependencies when building the project, so I've elected to drop 2.10 support with this PR.
- Cleans up `build.sbt` now that it seems like Scala 2.13 is supported on all projects.

### DIFF
--- a/.travis-jvmopts
+++ b/.travis-jvmopts
@@ -1,8 +1,8 @@
 # This is used to configure the sbt instance that Travis launches
 
--Xms2G
--Xmx2G
--Xss2M
+-Xms4G
+-Xmx4G
+-Xss4M
 -XX:MaxMetaspaceSize=1G
 -XX:+CMSClassUnloadingEnabled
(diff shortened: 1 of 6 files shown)


lloydmeta:
Thanks for this @a1kemist, and apologies for releasing later than I said. Just merged, and pushed a new release. It should be up on Maven Central in a bit.

PR #270, opened 2020-02-10
reactivemongo: Only depend on the bson-api, noshaded variant

avdv:
There is no need to pull in reactivemongo, the driver, since all we are
concerned with is the BSON API.

Also, starting with version 0.19.6, there are now two variants, a noshaded
version and a version which has Netty integrated into it for projects using
Netty themselves avoiding incompatibilities.

Since we do not know in advance which library is choosen in a project, make the
dependency a "provided" one.


lloydmeta:
Scratch that; currently on holiday and didn't bring the right laptop for publishing; will do when I get back.

PR #271, opened 2020-02-12
Circe 2.13 Release

ChristopherDavenport:


build.sbt
```diff
@@ -65,6 +65,7 @@ def thePlayJsonVersion(scalaVersion: String) =
 
 def theCirceVersion(scalaVersion: String) =
   CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, scalaMajor)) if scalaMajor >= 13 => "0.13.0"
```
lloydmeta: Looks like Circe 0.13 is also available for Scala 2.12, so we might as well bump that too?

ChristopherDavenport: Is the plan for you to release a new version with this then? That's a non-binary compatible change, whereas adding a new never published version is not. If that showed up in a patch version could easily break apps. I'm happy to switch it, just wanted to understand what the release strategy would be.

lloydmeta: Yeah, I intend to release a new version with this.

lloydmeta: Actually after merging, publishing, and re-reading this, I think it sounds like you were asking if I'll be releasing a new release *minor version* bump for the Circe integration lib?

If so... historically, I've not done this because I've kept the minor version in line with the main enumeratum core minor version and simply bumped the patch version when bumping integration lib versions.. no real good reason for it, just didn't have a good strategy for dealing with all the different versions. I'm open to suggestions for the future though.

PR #275, opened 2020-04-14
Find members in nested objects

mrdziuban:
This updates the `findValues` macro to recursively search the bodies of objects defined within the companion object, e.g.

```scala
sealed trait Thing extends EnumEntry
object Thing extends Enum[Thing] {
  object nested {
    // this object is found now
    case object Foo extends Thing

    object nested2 {
      // this is found too
      case object Bar extends Thing {

        // even this one is found
        case object Baz extends Thing
      }
    }
  }
}
```

Note: I expect CI will fail since [build.sbt references the stable version of enumeratum-macros](https://github.com/lloydmeta/enumeratum/blob/master/build.sbt#L186) which doesn't contain these changes. I tested locally by uncommenting the `dependsOn(macros)` line, but figured I shouldn't commit that. Let me know what the right workflow is for getting everything up to date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,13 +145,36 @@ Greeting.withNameLowercaseOnlyOption("hello")
 Note that by default, `findValues` will return a `Seq` with the enum members listed in written-order (relevant if you want to
 use the `indexOf` method).
 
+Enum members found in nested objects will be included by `findValues` as well, and will appear in the order they are
+written in the companion object, top to bottom. Note that enum members declared in traits or classes will *not* be
+discovered by `findValues`. For example:
+
+```scala
+sealed trait Nesting extends EnumEntry
+object Nesting extends Enum[Nesting] {
+  val values = findValues
+
+  case object Hello extends Nesting
+  object others {
+    case object GoodBye extends Nesting
+  }
+  case object Hi extends Nesting
+  class InnerClass {
+    case object NotFound extends Nesting
+  }
+}
+
+Nesting.values
+// => res0: scala.collection.immutable.IndexedSeq[Nesting] = Vector(Hello, GoodBye, Hi)
+```
+
 For an interactive demo, checkout this [Scastie snippet](https://scastie.scala-lang.org/lloydmeta/AMXiGvzkR0CD5sgWXiW4bg).
 
 ## More examples
 
 ### Enum
 
-Continuing from the enum declared in [the quick-start section](#usage):
+Continuing from the `Greeting` enum declared in [the quick-start section](#usage):
 
 ```scala
 import Greeting._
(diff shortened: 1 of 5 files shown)


lloydmeta:
Interesting. Not sure if there's any reason _not_ to do this. Let's see if CI fixes itself..

mrdziuban:
@lloydmeta I suppose it technically is a breaking change if any users were relying on the current behavior, so maybe a major version bump would be best if you're following semantic versioning. Would you like me to re-run the CI? I don't think it will pass without updating the macros version, but I'm happy to try

lloydmeta:
Sounds reasonable to me. 

Looks like CI is weird but between work and covid-19, I don't have time to take a closer look to see what's wrong (could be Travis? not sure). Will try to make some time to look into this when things are less busy :(

mrdziuban:
Understandable, I've gone ahead and published the core and macros projects in my fork to bintray so I can start using it.

lloydmeta:
Just took a look, and I think this fails some specs

https://travis-ci.org/github/lloydmeta/enumeratum/jobs/677665867#L2208

```
[info] EnumSpec:
2202[info] no values
2203[info] - should result in findValues finding nothing
2204[info] when not wrapped in another object
2205[info]   #values
2206[info]   - should contain objects
2207[info]   - should contain objects found in nested objects *** FAILED ***
2208[info]     Vector(Hello, GoodBye) was not equal to Vector(Hello, GoodBye, Hello, GoodBye, NestedInsideMember, Hello, GoodBye, Hello) (EnumSpec.scala:27)
```

mrdziuban:
Yes @lloydmeta that's the test I added. As I said in my original comment, it will fail on travis because  [the `core` project in `build.sbt` depends on the stable version of `macros`](https://github.com/lloydmeta/enumeratum/blob/master/build.sbt#L186), which obviously does not contain the changes in this pull request. If I comment out that line and uncomment [this one](https://github.com/lloydmeta/enumeratum/blob/master/build.sbt#L188) then all of the tests pass. I've just pushed e924537 with this change and [the CI build succeeds now](https://travis-ci.org/github/lloydmeta/enumeratum/jobs/678989803).

mrdziuban:
Yes, only objects defined within other objects (not classes or traits) will be found. I added a member within a class to  the`NestedObjectEnum` to prove this.

And the index order is preserved as well, I added some checks to the `indexOf` test to verify this. The [explicit order listed in the other test](https://github.com/lloydmeta/enumeratum/pull/275/files#diff-aca8bb75150a9fa937e2084fce69328fR28-R39) also proves it.

lloydmeta:
@mrdziuban Thanks for this, I've just pushed out a new release with this update.

mrdziuban:
Thanks @lloydmeta!

Issue #277, opened 2020-04-15
Doobie enums integration seems broken


FunctorPrototype:
scala 2.13.1
doobie 0.9.0
enumeratum 1.5.15
enumeratum-doobie 1.5.17
postgresql 12.2

It seems that using enums with doobie is broken, because ```ShirtSize``` example from readme throws 
```"org.postgresql.util.PSQLException: ERROR: column "shirt" is of type shirtsize but expression is of type character varying"```

The only workaround that I've found is
 ```CREATE CAST (character varying AS shirtsize) WITH INOUT AS ASSIGNMENT; ``` 
It does the trick

lloydmeta:
I personally don't use this integration and am not sure what the idiomatic usage here should be.

FunctorPrototype:
I've understood. The problem is that readme example intending usage of varchar column but I use postgres enum column

FunctorPrototype:
I'll create pr to documentation for more clear description of example

PR #278, opened 2020-04-16
Extended doobie example description

FunctorPrototype:
Due to misunderstanding that can appear [like here](https://github.com/lloydmeta/enumeratum/issues/277) I think it would be better to add schema db description. It will help to get entire picture of example

### DIFF
--- a/README.md
+++ b/README.md
@@ -1133,7 +1133,14 @@ libraryDependencies ++= Seq(
 ### Usage
 
 #### Enum
+If you need to store enum values in text column of following table
 
+```sql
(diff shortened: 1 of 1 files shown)


lloydmeta:
Thanks ! This LGTM.

Looks like CI is weird but between work and covid-19, I don't have time to take a closer look to see what's wrong (could be Travis? not sure). Will try to make some time to look into this when things are less busy :(

FunctorPrototype:
Ok, I'll make another PR with additional info about postgres enums. P.S. Yes, CI for readme changes is seems to be redundant

Issue #279, opened 2020-04-21
PlayInsensitiveJsonEnum: missing implicit compile error


blanchet4forte:
Hello,

When I try and use `PlayInsensitiveJsonEnum` I get a missing implicit writer error. When I change back to using `PlayJsonEnum` the compile error goes away.

The only thing I can think of is that `PlayInsensitiveJsonEnum` does not have the `contraJsonWrites` implicit def that `PlayJsonEnum` has.

lloydmeta:
Ah yeah, looks like it was missed in https://github.com/lloydmeta/enumeratum/pull/269.

PR #288, opened 2020-07-31
Play json & Argonaut sjs 1.x support

a1kemist:
## Purpose

+ Reinstates play-json js project
  - scala `2.12.x` and `2.13.x` => `2.9.0`
  - scala `2.11.x` => js still not supported (and unlikely in the future)
+ Reinstates argonaut js project
  - scala `2.12.x` and `2.13.x` => `6.3.0`
  - scala `2.11.x` => `6.2.5`
+ Fixes regression (#279) introduced in (#269)
  - Test cases added in play-json module that would have caught the regression
+ Bumps play to `2.8.0` for scala `2.12.x` and `2.13.x`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Enumeratum is published for Scala 2.11.x, 2.12.x, 2.13.x as well as ScalaJS.
 Integrations are available for:
 
 - [Play](https://www.playframework.com/): JVM only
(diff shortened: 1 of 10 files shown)


lloydmeta:
@a1kemist thanks so much for doing this. I've just pushed new releases to Maven Central for this.

PR #291, opened 2020-09-02
Upgrade ReactiveMongo to 1.0.0

cchantep:


build.sbt
```diff
@@ -13,7 +13,7 @@ lazy val scalaTestVersion  = "3.1.1"
 lazy val scalacheckVersion = "1.14.3"
 
 // Library versions
-lazy val reactiveMongoVersion = "1.0.0-noshaded-rc.2"
```
lloydmeta, line 16: Out of curiosity; this switches from `noshaded`; is there a reason here for that? Not sure what's recommended in this case and wanted to check + learn.

lloydmeta (commented):
LGTM, just left a question re: `noshaded`

build.sbt
cchantep: Noshaded is only for advanced user who want to provide netty by themselves. Not recommended for dev who are not sure and have hard time to adjust version requirements.

lloydmeta (approved):
LGTM

lloydmeta:
Thanks for this @cchantep Just pushed a release with this bump :)

PR #302, opened 2021-02-23
Play-JSON KeyReads & KeyWrites

cchantep:
Required to use enum in `Map` with Play-JSON


lloydmeta (approved):
LGTM

lloydmeta:
Thanks, pushed to Maven central as 1.6.2

blanchet4forte:
Hmm. Not sure what I'm doing wrong. Getting this error:
```
No instance of play.api.libs.json.Format is available for scala.collection.immutable.Map[io.rallynetwork.rnb.shared.ReportKey, java.lang.String] in the implicit scope
```

It works in any other context. Just not in `Map`.

Issue #309, opened 2021-03-24
Improve the usability of error messages in UrlBinder


sapka12:
https://github.com/lloydmeta/enumeratum/blob/b6548e27034f57597b0cf665d0ba836c295a2dbe/enumeratum-play/src/main/scala/enumeratum/UrlBinders.scala#L84

In the case of None, there is only a reference for the given UrlBinder. It would be more useful to return the enum itself.

`${this.toString}` should probably be `${enum.toString}` 

(I can send a PR is this change is acceptable)

lloydmeta:
Thanks for this! Just pushed `1.6.3` to Maven Central that has this fix.

PR #310, opened 2021-03-25
change UrlBinder error message to use the enum

sapka12:
resolving the following issue: https://github.com/lloydmeta/enumeratum/issues/309


lloydmeta (approved):
Makes sense to me. Thanks.

PR #319, opened 2021-06-16
Add Enum.extraNamesToValuesMap method

agustafson:
This is a PR for issue #321 

In order to map legacy string values to an `Enum`, I added a `Enum.extraNamesToValuesMap` which allows the user to define extra mappings.

Example from test:
```scala
sealed trait Result extends EnumEntry
object Result extends Enum[Result] {
  case object Good extends Result
  case object Bad  extends Result

  override val extraNamesToValuesMap: Map[String, Result] = Map[String, Result]("Ok" -> Good)

  override val values = findValues
}
```

In the case above, `Result.withName("Ok")` will return `Result.Good`.


lloydmeta (approved):
LGTM, just pending some things I want to fix on CI before merging.

lloydmeta:
Thanks for this. Just released `1.7.0` with this.

agustafson:
Great! Thanks so much for the speedy merge and release @lloydmeta - much appreciated!

And it's a great library too btw.

PR #331, opened 2021-09-15
Initialization-order problem example.

rulle-io:
Attempt to define a problem with a failed unit-test.
I am not sure if anything can be fixed in enumeratum code.

The problem is defined in [Scala FAQ: "Why is my abstract or overridden val null?"](https://docs.scala-lang.org/tutorials/FAQ/initialization-order.html)


lloydmeta:
This will happen even without enumeratum if you expand the `val` to the Seq; I'm going to close this because this is just a standard Scala init-order issue,

PR #335, opened 2021-11-25
Generate Codec instance from Decoder and Encoder

regiskuckaertz:
Hello 👋 

Big fan here. We have been adding these codecs instances manually so far and I thought they may as well be defined in this library instead.

Let me know what you think! Happy to add a test if you think it is useful.

### DIFF
--- a/enumeratum-circe/src/main/scala/enumeratum/Circe.scala
+++ b/enumeratum-circe/src/main/scala/enumeratum/Circe.scala
@@ -2,7 +2,7 @@ package enumeratum
 
 import cats.syntax.either._
 import io.circe.Decoder.Result
-import io.circe.{Encoder, Decoder, Json, HCursor, DecodingFailure, KeyEncoder, KeyDecoder}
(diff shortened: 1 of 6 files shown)


lloydmeta:
I'm not opposed to this in principal, but I think this is a breaking change for situations where an implicit `Encoder[A]` or `Decoder[A]` is expected, since the new implicit `Codec[A]`-generating method would make those ambiguous, as per this compilation failure; https://github.com/lloydmeta/enumeratum/runs/4330281650?check_suite_focus=true#step:5:397.

I _feel_ like making this new `Codec[A]` generator either non-implicit, or, implicit-but-separately-imported (maybe in a separate object) is more in line with the spirit of how `Codec` was introduced to Circe in https://github.com/circe/circe/pull/1151.. WDYT?

regiskuckaertz:
Yes! You are 💯 correct and I should have thought about that. I feel like the second solution is slightly better, as it will reduce user-land code. Are you happy with `CirceCodecEnum` and `CirceCodecValueEnum`?

Issue #329, opened 2021-08-13
 Coloncase EnumEntry


lderks:
Hi, 

Could you please add support for a colon (:) `EnumEntry` as in:

`Coloncase => colon:case`
`UpperColoncase => UPPER:COLON:CASE`
`CapitalColoncase => Capital:Colon:Case`

stsatlantis:
@lderks  #342 solves your problem, if it won't be merged you can still copy the solution and use it in your project.

PR #342, opened 2022-02-19
#329 add colon case

stsatlantis:
Solves #329 

To be honest I don't know if this PR even makes sense, because it can be done on the project level as well.


lloydmeta:
Thanks, but not really looking to expand/add more of these helper classes. I've found them hard to maintain and reason about, with a number of people asking how things work when you stack things together.

stsatlantis:
Thanks for the response, makes sense :)

PR #360, opened 2023-03-01
Update slick (3.4.1)

kijuky:
# Sumary

Update slick 3.4.1

# Background

I would like to use this library with play + slick. [The latest play-slick](https://mvnrepository.com/artifact/com.typesafe.play/play-slick_2.13/5.1.0) conforms to slick 3.4.1 and conflicts with the version of slick that enumeratum-slick depends on.


lloydmeta:
Thanks for this. Just pushed this to Maven central as `enumeratum-slick` 1.7.3.

PR #364, opened 2023-06-09
chore: bump json4s and publish also for Scala 3

ThijsBroersen:


ThijsBroersen:
still blocked by json4s bug it seems: https://github.com/json4s/json4s/issues/1035

PR #365, opened 2023-06-21
Upgrade doobie to 1.0.0-RC4

jatcwang:


Awethon:
@lloydmeta could you please take a look?

lloydmeta (commented):
Overall LGTM, but I think the build is failing, can you please look into that?

limansky:
It looks like reactive mongo dependency should be a RC7, without snapshot. I've created PR #366  to fix it, and update to latest RC.

jatcwang:
1.0.0-RC4 is using Scala 3.3.0 so I think we may need to upgrade scala 3 ver here too.

lloydmeta:
Closing this because it was handled via https://github.com/lloydmeta/enumeratum/pull/370

PR #371, opened 2023-07-19
Update scala 3 macro to support a hierarchy of sealed traits

### LINKED ISSUE #368: Scala3: findValues fails to compile in nested sealed trait hierarchy
Example: 
```scala
import enumeratum.values.StringEnum
import enumeratum.values.StringEnumEntry

sealed trait FirstLevel extends StringEnumEntry

object FirstLevel extends StringEnum[FirstLevel] {

  case object One extends FirstLevel {
    val value = "one"
  }
  case object Two extends FirstLevel {
    val value = "two"
  }

  sealed trait SecondLevel extends FirstLevel

  case object Three extends SecondLevel {
    val value = "three"
  }

  override def values: IndexedSeq[FirstLevel] = findValues

}
```

https://scastie.scala-lang.org/qpDNUwqAT4m4ovmn4EySgg

Error:
Exception occurred while executing macro expansion.
scala.MatchError: Type.of[...] (of class scala.quoted.runtime.impl.TypeImpl)

coreywoodfield:
This fixes #368. I wasn't sure how to add/update tests, but I did test it in a sample project locally, and it works as expected. It finds all the instances of the base `sealed trait` and all the instances of the inheriting `sealed trait`

### DIFF
--- a/enumeratum-core/src/test/scala/enumeratum/values/ValueEnumSpec.scala
+++ b/enumeratum-core/src/test/scala/enumeratum/values/ValueEnumSpec.scala
@@ -135,6 +135,22 @@ class ValueEnumSpec extends AnyFunSpec with Matchers with ValueEnumHelpers {
        """ should (compile)
       }
 
+      it("should compile when there is a hierarchy of sealed traits") {
+        """
+        sealed abstract class Top(val value: Int) extends IntEnumEntry
+        sealed trait Middle extends Top
+
+        case object Top extends IntEnum[Top] {
+          case object One extends Top(1)
+          case object Two extends Top(2)
+          case object Three extends Top(3) with Middle
+          case object Four extends Top(4) with Middle
+
+          val values = findValues
+        }
+        """ should compile
+      }
+
       it("should fail to compile when there are non literal values") {
         """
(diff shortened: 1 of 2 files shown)


lloydmeta (commented):
Thanks for taking a stab at this.

Regarding tests, it might be good to add something to

https://github.com/lloydmeta/enumeratum/blob/74a7b36853cc52df832ceb56279a1b67f835fb40/enumeratum-core/src/test/scala/enumeratum/values/ValueEnumSpec.scala#L10

possibly by adding a sealed trait enum definition in `enumeratum-core/src/test/scala/enumeratum/values`

coreywoodfield:
I added a test. It fails if I run `coreJVM / test` as-is and passes if I run `coreJVM / test` after settings `useLocalVersion` in `build.sbt` to `true`

lloydmeta (approved):
LGTM thanks ! I'll merge this once I have some time to push out releases.

Re: having to set the "use local version" flag: that's expected :)

lloydmeta:
Pulled this branch and merged in master locally and ran some basic tests...seems to be good so merging now.

lloydmeta:
Thanks for this btw, it has been released as part of https://github.com/lloydmeta/enumeratum/releases/tag/macros-1.6.4

PR #379, opened 2023-10-30
Bump play-json

markarasev:


lloydmeta (approved):
LGTM. Just waiting for the weekend to merge + release.

gaeljw:
Shouldn't we bump to `org.playframework:play-json:3.0.0` directly?

The artifact groupId is renamed but the content is the same (https://github.com/playframework/play-json/compare/2.10.1...3.0.0).

lloydmeta:
> Shouldn't we bump to `org.playframework:play-json:3.0.0` directly?
> 
> The artifact groupId is renamed but the content is the same ([playframework/play-json@2.10.1...3.0.0](https://github.com/playframework/play-json/compare/2.10.1...3.0.0)).

That could be bigger + separate PR that would be very welcome indeed.

gaeljw:
I'll provide the PR today then :)

PR #381, opened 2023-11-04
Upgrade Play & Play JSON to 3.0

gaeljw:
Supersedes https://github.com/lloydmeta/enumeratum/pull/380

- Play JSON 3.0 is identical to Play JSON 2.10.1 https://github.com/playframework/play-json/compare/2.10.1...3.0.0, it's only a groupId renaming of the artifact (`com.typesafe.play` -> `org.playframework`)
- Play 3.0 brings new features/bugfixes, drop support for Scala 2.12 and uses Pekko instead of Akka (though it seems to me that enumeratum code is unrelated to Pekko/Akka)

As discussed in https://github.com/lloydmeta/enumeratum/pull/380, Play 2.x support is discontinued in favour of Play 3.x unless there is a strong demand for it.

Closes https://github.com/lloydmeta/enumeratum/issues/382


build.sbt
```diff
@@ -320,20 +300,9 @@ lazy val enumeratumPlay = Project(id = "enumeratum-play", base = file("enumeratu
...
-          .exclude("org.scala-lang.modules", "*")
-          .exclude("com.typesafe.play", "play-json_2.13")
-          .cross(CrossVersion.for3Use2_13)
-      } else {
-        dep
-      }
-    },
-    libraryDependencies += scalaTestPlay(scalaVersion.value),
+    crossScalaVersions := Seq(scala_2_13Version, scala_3Version),
+    libraryDependencies += ("org.playframework" %% "play" % "3.0.0").exclude("org.scala-lang.modules", "*"),
```
gaeljw, line 304: @lloydmeta It's not clear to me why the `exclude("org.scala-lang.modules", "*")` is present, I kept it in doubt but maybe it's not needed anymore?

build.sbt
```diff
@@ -319,21 +299,13 @@ lazy val enumeratumPlay = Project(id = "enumeratum-play", base = file("enumeratu
...
-        dep
-      }
-    },
-    libraryDependencies += scalaTestPlay(scalaVersion.value),
+    version := Versions.Core.head,
+    // Play do not support 2.12 (default from common settings)
+    scalaVersion                                := scala_2_13Version,
+    crossScalaVersions                          := Seq(scala_2_13Version, scala_3Version),
+    libraryDependencies += ("org.playframework" %% "play" % "3.0.0")
+      .exclude("org.scala-lang.modules", "*"),
```
gaeljw, line 307: @lloydmeta It's not clear to me why the `exclude("org.scala-lang.modules", "*")` is present, I kept it in doubt but maybe it's not needed anymore?

gaeljw:
I don't understand why 2.12 build fails but for sure it's related to my attempt to not build enumeratum-play for 2.12 anymore. Will investigate a bit more tomorrow. Unless you've got any idea.

gaeljw:
> I don't understand why 2.12 build fails but for sure it's related to my attempt to not build enumeratum-play for 2.12 anymore. Will investigate a bit more tomorrow. Unless you've got any idea.

Wow.. this was tricky but I think I got it: from what I understand, the commands used in CI were not actually doing what you'd think. The `sbt ++<version> cmd1 cmd2` was running only cmd1 with the desired Scala version, cmd2 was running on all Scala versions. Sounds a bit weird, not sure I understood everything to be honest.

This is the reason I had to make changes to CI:
- actually run all commands with the desired Scala version
- use Scala 2.13 for `scalafmtCheck` and coverage reports (otherwise we lose coverage for enumeratum-play which is not available in 2.12)
- fix style so that `scalafmt` (on compile) runs without error with Scala 3 (it was failing on a macro, not sure why, just disabled it)

I've done several commits but happy to move these unrelated changes to another PR first if you prefer to.

lloydmeta (approved):
LGTM. Just holding on merging until I get time to also release.

lloydmeta:
Thanks for this. Just pushed 1.8.0 of the Play integration libs to Maven central based on this.

gaeljw:
Great news :)

PR #385, opened 2023-11-21
Update to doobie 1.0.0-RC5

limansky:


lloydmeta:
Oof. Do things work if you simply exclude the dependency when adding the enumeratum-doobie lib as a dependency? 

If that works, great. Otherwise I'm going to probably think twice about RC or M dependent versions of libs.

kovacshuni:
Release please

lloydmeta:
> Release please

@kovacshuni does what I described in https://github.com/lloydmeta/enumeratum/pull/385#issuecomment-1836014428 work for you?

limansky:
@lloydmeta I suppose there is no guarantee what the RC4 and RC5 are fully binary compatible. So if it works for some case it doesn't proof that it will work for another.

lloydmeta:
Fair enough, thanks @limansky . Just pushed to Maven Central.

Issue #300, opened 2021-02-03
Scala 3: What should be done


lloydmeta:
Looking at Scala 3, the [enumerations](https://dotty.epfl.ch/docs/reference/enums/enums.html) built into the language look pretty good.

The questions are

1. What should `enumeratum` look like in a Scala 3 world?
   * What problems are we trying to solve? The problems with Scala 2 enums were pretty obvious, which led to the core of this lib

catostrophe:
The first step forward Scala 3 for any real-world project is cross-compilation between Scala 2.13/Scala 3. Most of the OSS libs added Scala 3 to their cross-builds.

IMO, this will be a status quo for the next few years. Only then we'll drop Scala 2 support and migrate to the new enums.

Thus, the Enumeratum macros should be ported.

hmemcpy:
My 2c: we use Enumeratum mostly to model things that map to postgres enums, with a nice benefit of being able to have safe `withName*` overloads for decoding.

I'm not sure if Scala 3 enums has support for this...

lloydmeta:
> The first step towards Scala 3 for any real-world project is cross-compilation between Scala 2.13/Scala 3. Most of the OSS libs added Scala 3 to their cross-builds.
> 
> IMO, this will be a status quo for the next few years. Only then we'll drop Scala 2 support and migrate to the new enums.
> 
> Thus, the Enumeratum macros should be ported.

Fair point; have you checked that the macro(s) in its current form can be ported? A PR would be welcome to start concrete discussions as well.

lloydmeta:
> My 2c: we use Enumeratum mostly to model things that map to postgres enums, with a nice benefit of being able to have safe `withName*` overloads for decoding.
> 
> I'm not sure if Scala 3 enums has support for this...

I *think* there is a way of having custom names, based on this doc https://github.com/dotty-staging/dotty/blob/master/docs/docs/reference/enums/desugarEnums.md#translation-of-enums-with-singleton-cases

note:
@lloydmeta I think you were looking for `valueOf`: [scastie session](https://scastie.scala-lang.org/kMIypfZ3RWerLBySmzjf6A)

lloydmeta:
@note Hmmm I think that's almost it, but how do we make the second print resolve `Red` from the custom name `"red"`?

note:
[...]

PR #395, opened 2024-05-21
Mix scala 2 and 3 macros

gregghz:


lloydmeta:
Thanks for the PR but can you please add a description for this that explains why we want this?

gregghz:
@lloydmeta ah, thanks for taking a look. I've updated the description to clarify what's going on.

gregghz:
@lloydmeta thanks for your patience with this. I just updated things again to hopefully resolve the build failures. I can't remember, were you manually triggering builds?

lloydmeta:
Yea... I think I set something up once that requires me to approve the CI runs 🤦‍♂️

.github/workflows/ci.yml
```diff
@@ -13,9 +13,9 @@ jobs:
           - java: 11
             scala: 2.12.18
           - java: 11
-            scala: 2.13.11
+            scala: 2.13.13
           - java: 11
-            scala: 3.3.0
+            scala: 3.4.0
```
lloydmeta, line 18: This is not an LTS release. Let's revert this.

.scalafmt.conf
```diff
@@ -1,4 +1,4 @@
-version = 3.7.8
+version = 3.7.17
```
lloydmeta, line 1: Was this needed?

build.sbt
```diff
@@ -3,13 +3,13 @@ import sbtbuildinfo.BuildInfoPlugin.autoImport._
 import sbtcrossproject.CrossPlugin.autoImport.{CrossType, crossProject}
 
 lazy val scala_2_12Version = "2.12.18"
-lazy val scala_2_13Version = "2.13.11"
-lazy val scala_3Version    = "3.3.0"
+lazy val scala_2_13Version = "2.13.13"
+lazy val scala_3Version    = "3.4.0"
```
lloydmeta, line 7: Please revert.

build.sbt
```diff
@@ -108,11 +108,14 @@ lazy val macros = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     name    := "enumeratum-macros",
     version := Versions.Macros.head,
     crossScalaVersions := scalaVersionsAll, // eventually move this to aggregateProject once more 2.13 libs are out
-    libraryDependencies += {
+    libraryDependencies ++= {
       if (scalaBinaryVersion.value == "3") {
-        "org.scala-lang" %% "scala3-compiler" % scalaVersion.value % Provided
+        Seq(
+          "org.scala-lang" %% "scala3-compiler" % scalaVersion.value % Provided,
+          "org.scala-lang" % "scala-reflect" % scala_2_13Version,
```
lloydmeta, line 115: Curious if there's a way to make this additional dependency
* Contingent on actually using the mix mode macro
* Only a compile time constraint?

build.sbt
```diff
@@ -685,7 +688,7 @@ lazy val compilerSettings = Seq(
 
     val base = {
       if (scalaBinaryVersion.value == "3") {
-        minimal :+ "-deprecation"
+        minimal :+ "-Wconf:cat=deprecation:s"
```
lloydmeta, line 691: Curious why this was needed?

enumeratum-cats/src/main/scala/enumeratum/values/Cats.scala
```diff
@@ -7,7 +7,7 @@ object Cats {
   /** Builds an `Eq` instance which differentiates all enum values as it's based on universal
     * equals.
     */
-  def eqForEnum[A <: ValueEnumEntry[_]]: Eq[A] = Eq.fromUniversalEquals[A]
+  def eqForEnum[A <: ValueEnumEntry[?]]: Eq[A] = Eq.fromUniversalEquals[A]
```
lloydmeta, line 10: Was this needed?

enumeratum-core/src/main/scala-3/enumeratum/EnumCompat.scala
```diff
@@ -8,7 +11,8 @@ private[enumeratum] trait EnumCompat[A <: EnumEntry] { _enum: Enum[A] =>
     * You will want to use this in some way to implement your [[values]] method. In fact, if you
     * aren't using this method... why are you even bothering with this lib?
     */
-  inline def findValues: IndexedSeq[A] = ${ EnumMacros.findValuesImpl[A] }
+  protected inline def findValues: IndexedSeq[A] = ${ EnumMacros.findValuesImpl[A] }
```
lloydmeta, line 14: Curious why this is now protected.

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
```diff
@@ -19,7 +19,7 @@ trait EnumEntry {
     */
   def entryName: String = stableEntryName
 
-  private[this] lazy val stableEntryName: String = toString
+  private lazy val stableEntryName: String = toString
```
lloydmeta, line 22: Was this (and below) removals of `this` needed?

enumeratum-play/src/main/scala/enumeratum/values/Forms.scala
```diff
@@ -9,7 +9,7 @@ import play.api.data.{FormError, Mapping, Forms => PlayForms}
   */
 object Forms extends FormsCompat {
 
-  protected[this] def formatter[ValueType, EntryType <: ValueEnumEntry[
+  protected def formatter[ValueType, EntryType <: ValueEnumEntry[
```
lloydmeta, line 12: ?

macros/src/main/scala/enumeratum/compat/EnumMacros.scala
```diff
@@ -10,7 +11,7 @@ object EnumMacros {
 
   /** Finds any [A] in the current scope and returns an expression for a list of them
     */
-  def findValuesImpl[A: c.WeakTypeTag](c: Context): c.Expr[IndexedSeq[A]] = {
+  def findValuesImpl[A: c.WeakTypeTag](c: Context): c.Tree = {
```
lloydmeta, line 14: Was this needed? Feels like a step backwards to lose type info, so if it's not required, please revert.

macros/src/main/scala/enumeratum/compat/ValueEnumMacros.scala
lloydmeta, line 1: Same question regarding type change for everything here.

lloydmeta (changes requested):
Thanks for doing this. 

There are quite a lot of changes that do not _feel_ necessary, some of which are IMO a step backwards. My request would be to remove any changes that are not absolutely necessary.

enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
gregghz: the scala 3 build fails without removing these (deprecated feature)

.github/workflows/ci.yml
gregghz: Unfortunately, this PR won't work with scala < 3.4 because of https://github.com/scala/scala3/issues/16630

They mentioned backporting to 3.3.1-RC1 but seems like that never happened. It was also mentioned briefly here: https://github.com/lloydmeta/enumeratum/issues/300#issuecomment-1374313401

lloydmeta: Understood. That is a blocker indeed.

gregghz: Makes sense. I tested this out using scala 3.3.4-RC4 and it does build succesfully with that version. So once 3.3.4 gets released I'll udpate this PR.

.scalafmt.conf
gregghz: Seems like no, I'll revert.

build.sbt
gregghz: Unfortunately I think only the consumer of the library knows if the mixed mode macro is being used, from here, sbt only knows that it's a scala3 build.

I believe we could mark this as `Provided` too, but then all consumers would need to explicitly provide it. Marking it provided here means I also need to add it as Provided in `core`. It would only need to be added as a normal/runtime dep if the project was using the mixed mode because I think this is only needed for the scala 2 portion of the build.

build.sbt
gregghz:
```
[error] -- Error: /home/gregg/lucid/enumeratum/macros/src/main/scala/enumeratum/compat/EnumMacros.scala:109:32 
[error] 109 |        val enclosingModule = c.enclosingClass match {
[error]     |                              ^^^^^^^^^^^^^^^^
[error]     |method enclosingClass in trait Enclosures is deprecated since 2.11.0: c.enclosingTree-style APIs are now deprecated; consult the scaladoc for more information
```

and a few more identical errors

build.sbt
gregghz: @lloydmeta let me know if you have a preference here.

build.sbt
gregghz: reverted

.scalafmt.conf
gregghz: reverted

.github/workflows/ci.yml
gregghz: this is fixed now. 3.3.4 was released.

.scalafmt.conf
gregghz: I changed this to 3.7.10 because without that update, in sbt `++3.3.4 test:compile` fails with:

```
[error] (coreJS / Compile / scalafmt) org.scalafmt.sbt.ScalafmtSbtReporter$ScalafmtSbtError: scalafmt: /home/gregg/lucid/enumeratum/enumeratum-core/src/main/scala-3/enumeratum/EnumCompat.scala:14: error: [dialect scala3] pattern must be a value or have parens: EnumMacros.findValuesImpl[A]
[error]   protected inline def findValues: IndexedSeq[A] = ${ EnumMacros.findValuesImpl[A] }
[error]                                                                                    ^ [/home/gregg/lucid/enumeratum/enumeratum-core/src/main/scala-3/enumeratum/EnumCompat.scala]
```
there and a few other places.

enumeratum-core/src/main/scala-3/enumeratum/EnumCompat.scala
gregghz: I'm not sure. I reverted it.

enumeratum-play/src/main/scala/enumeratum/values/Forms.scala
jadenPete: This change has been reverted.

enumeratum-cats/src/main/scala/enumeratum/values/Cats.scala
jadenPete: This change has been reverted.

build.sbt
lloydmeta: Still having a think about this - has there been any better solution than the ones presented here? Adding another runtime dependency is not great.

Personally I think it's fine if the experience is subpar for  "mixed mode" but not for Scala 2 or 3 users.

lloydmeta:
Thanks for checking up. I'm happy to close it unless there's much more demand for it.

PR #397, opened 2024-06-19
draft: don't check for the correct param type for empty tree

martijnhoekstra:
`p.tree` ends up being the empty tree, so it makes no sense to test for it.

Can we just do this? I don't understand the code around it well enough to be sure.


martijnhoekstra:
This needs more work

PR #398, opened 2024-06-21
allow value constructor param to be any constructor parameter, rather than only the first

martijnhoekstra:
simplify and correct finding the constructor param index for the value param.

As a follow-up, I want to double-check whether we need to require -Yretain-trees.

Some of the boyscouting here is a matter of taste, if it's not to yours, I'm happy to revert.

### DIFF
--- a/macros/src/main/scala-3/enumeratum/ValueEnumMacros.scala
+++ b/macros/src/main/scala-3/enumeratum/ValueEnumMacros.scala
@@ -171,14 +171,14 @@ In SBT settings:
             vof <- Expr.summon[ValueOf[h]]
             constValue <- htpr.typeSymbol.tree match {
               case ClassDef(_, _, parents, _, statements) => {
-                val fromCtor = parents
-                  .collectFirst {
-                    case Apply(Select(New(id), _), args) if id.tpe <:< repr               => args
-                    case Apply(TypeApply(Select(New(id), _), _), args) if id.tpe <:< repr => args
-                  }
-                  .flatMap(_.lift(valueParamIndex.getOrElse(-1)).collect { case ConstVal(const) =>
-                    const
-                  })
+                val fromCtor = valueParamIndex.flatMap { ix =>
+                  parents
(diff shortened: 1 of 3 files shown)


martijnhoekstra:
all the stuff in the build files shouldn't have come along

lloydmeta:
Thanks for this. I will definitely review this once I have some spare time 🙇‍♂️

macros/src/main/scala-3/enumeratum/ValueEnumMacros.scala
```diff
@@ -248,7 +218,7 @@ In SBT settings:
               }
               .mkString(", ")
 
-            Left(s"Values for ${valueField.name} are not discriminated subtypes: ${details}")
+            Left(s"Values for the `value` field field are not discriminated subtypes: ${details}")
```
lloydmeta, line 221: Can you please change this to 

```
Values for value are not...
```

?

lloydmeta (commented):
LGTM. Left a question and two suggestions for reverting error messages.

++ on the simplification here.

macros/src/main/scala-3/enumeratum/ValueEnumMacros.scala
```diff
@@ -220,7 +191,7 @@ In SBT settings:
 
             case None =>
               report.errorAndAbort(
-                s"Fails to check value entry ${htpr.show} for enum ${repr.show}"
+                s"Failed to check value entry ${htpr.show} for enum ${repr.show}"
```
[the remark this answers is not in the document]
martijnhoekstra: done

macros/src/main/scala-3/enumeratum/ValueEnumMacros.scala
martijnhoekstra: done

lloydmeta (approved):
LGTM. I'm holding off on merging until I have time to release a new version.

PR #413, opened 2025-02-10
Doobie 1.0.0-rc8

limansky:
Hi,

  I've updated code to use doobie-1.0.0-RC7, and also added some dependencies and tools updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,11 @@ jobs:
       matrix:
         include:
           - java: 11
-            scala: 2.12.18
+            scala: 2.12.20
           - java: 11
-            scala: 2.13.11
+            scala: 2.13.16
           - java: 11
-            scala: 3.3.0
+            scala: 3.3.5
     runs-on: ubuntu-latest
     env:
       SCALAJS_TEST_OPT: full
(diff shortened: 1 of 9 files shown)


lloydmeta (approved):
LGTM; I'm going to leave this open until I have time to publish.

limansky:
@lloydmeta doobie 1.0.0-RC8 was released last week, so I've updated this PR.

Dichotomia:
Hello.
When this could be released ?
We are blocked to our doobie upgrades because of this.

Thanks for all the hard work.

lloydmeta:
Just haven't had time. And, given that _2_ RCs came out in a few weeks, and RCs have been coming out for what, the last 3 years at least, I'm wondering if this is the right repo for this integration , or whether someone with more time is a better steward.

limansky:
@lloydmeta I can try to help with github actions scripting for release versions if you want me to help. I just think in this case it would be better to have same release cycle for all enumeratum modules (what can be a good idea btw).

lloydmeta:
> @lloydmeta I can try to help with github actions scripting for release versions if you want me to help. I just think in this case it would be better to have same release cycle for all enumeratum modules (what can be a good idea btw).

Sure please give it a stab if you're willing (maybe in a separate PR though).

lloydmeta:
Thanks for this @limansky. Released as enumeratum-doobie 1.7.8.

Issue #415, opened 2025-03-12
EnumEntry.uncapitalize is no longer safe to use .tail on empty String since Scala 2.13.16


firehooper:
The EnumEntry.uncapitalize is no longer safe to use .tail on empty String.
Use case: Sometimes we have an enum representing a empty or unknown with an empty string key.  

https://github.com/scala/bug/issues/13026

PR #416, opened 2025-03-12
415 EnumEntry.uncapitalize is no longer safe to use .tail on empty String since Scala 2.13.16

firehooper:
https://github.com/lloydmeta/enumeratum/issues/415

### DIFF
--- a/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
+++ b/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala
@@ -42,11 +42,11 @@ object EnumEntry {
 
   private def capitalise(str: String): String = {
     if (str.isEmpty) str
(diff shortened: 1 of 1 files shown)


lloydmeta (commented):
How is this unsafe in practice when we are already doing a "isEmpty" check right above it?

firehooper:
@lloydmeta for capitalize yes it is safe, but for uncapitalize we had a use case where using json4s and an empty string representation would throw.  I just changed them both to make them the same.  Avoiding the helper traits also are a valid workaround.

lloydmeta (approved):
LGTM.

lloydmeta:
@firehooper we require verified signatures on all commits made to this repo. Can you please address that so we can merge?

firehooper:
@lloydmeta signed the commit.  Hopefully correct.  Thanks for this awesome library.

Issue #424, opened 2025-05-14
[Scala 3] Type-parametrized enums with intermediate hierarchies not picked by findValues


Krever:
I found the behavior below during migration to Scala 3 and I couldn't find a workaround unfortunately (hence it's blocking the migration). 

```
import enumeratum.*

sealed trait Foo2[T] extends EnumEntry with Serializable

object Foo2 extends Enum[Foo2[Unit]] {
  sealed trait Bar[T] extends Foo2[T]
  
  case object A extends Bar[Unit]
  case object B extends Foo2[Unit]
  lazy val values: IndexedSeq[Foo2[Unit]] = findValues
}

println(Foo2.values)
// Vector(B)
```
Scastie: https://scastie.scala-lang.org/w9VdeWnJTH2O1jJwAKunRQ

So we define an intermediate node in the hierarchy (`Bar`) and its subtypes are not found by the macros. This was working fine in scala 2. ([Scastie](https://scastie.scala-lang.org/Wh8WAxBSQ7q0blY77TcZVw))

It might sound a bit artificial, but we have an essential and quite complex piece of domain modeling that looks like follows (simplified)

```
sealed trait Account[A] extends EnumEntry
object Account {
  sealed trait Asset[A] extends Account[A]
  sealed trait Liability[A] extends Account[A]
}
object UserAccounts extends Enum[Account[User]] { ... }
object CompanyAccounts extends Enum[Account[Company]] { ... }
```

Krever:
I lack the required macro expertise, but I think I pinpointed the issue. [Here](https://github.com/lloydmeta/enumeratum/blob/master/macros/src/main/scala-3/enumeratum/EnumMacros.scala#L156) we check if the child is of an entry type (`Foo2[Unit]` in my example). Obviously `Bar[T]` is not, hence its subtypes are not being considered. 
Unfortunately I don't know yet what a proper way to fix this is.

lloydmeta:
Yeah, I'm not actually sure I understand what is missing in Scala 3, and why #426 is needed (and more importantly why it is _not_ something that would poke a hole through things). 

https://scastie.scala-lang.org/tAL31BP3TASif48PIzohgQ

```scala
import enumeratum.*

sealed trait Foo2[T] extends EnumEntry with Serializable
sealed trait Bar[T] extends Foo2[T]

object Foo2 extends Enum[Foo2[?]] {
  
  
  case object A extends Bar[Int]
  case object B extends Foo2[Unit]
[...]

PR #426, opened 2025-05-20
Allow finding values inheriting from a generic intermediary subtype

Krever:
Aims to solve https://github.com/lloydmeta/enumeratum/issues/424. 

Original behavior: search among subclasses only of the enum type
New behavior: search among all subclasses, taking only those inheriting from the enum type

I’ve put `NestedGenericEnum` inside the test because I think it's more obvious/readable that way but I'm totally fine to move it into Models/scala to keep consistency.

### DIFF
--- a/enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala
+++ b/enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala
@@ -43,6 +43,23 @@ class EnumSpec extends AnyFunSpec with Matchers with EnumSpecCompat {
 
         Word.values should be(IndexedSeq(Word.Hello, Word.Hi))
       }
+
+      it("should contain instance of generic subclass") {
+        sealed trait NestedGenericEnum[T] extends EnumEntry with Serializable
+
+        object NestedGenericEnum extends Enum[NestedGenericEnum[Unit]] {
+          sealed trait Intermediary[T] extends NestedGenericEnum[T]
+
+          case object A extends Intermediary[Unit]
+          case object B extends NestedGenericEnum[Unit]
+          val values = findValues
+        }
+
+        NestedGenericEnum.values should contain theSameElementsAs Seq(
+          NestedGenericEnum.A,
+          NestedGenericEnum.B
+        )
+      }
     }
 
     describe("#withName") {
--- a/enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala
+++ b/enumeratum-core/src/test/scala/enumeratum/EnumSpec.scala
@@ -60,6 +60,21 @@ class EnumSpec extends AnyFunSpec with Matchers with EnumSpecCompat {
           NestedGenericEnum.B
         )
       }
+
+      it("should contain exactly one instance even if it inherits two subclasses") {
+        sealed trait MyEnum extends EnumEntry with Serializable
+        object MyEnum extends Enum[MyEnum] {
+          sealed trait Intermediary1 extends MyEnum
+          sealed trait Intermediary2 extends MyEnum
+
+          case object A extends Intermediary1 with Intermediary2
+          val values = findValues
+        }
+
+        MyEnum.values should contain theSameElementsAs Seq(
+          MyEnum.A
+        )
+      }
     }
 
     describe("#withName") {
--- a/macros/src/main/scala-3/enumeratum/EnumMacros.scala
+++ b/macros/src/main/scala-3/enumeratum/EnumMacros.scala
@@ -167,7 +167,7 @@ object EnumMacros:
 
     tpr.classSymbol
       .flatMap { cls =>
-        val types = subclasses(cls.children.map(_.tree), Nil)
+        val types = subclasses(cls.children.map(_.tree), Nil).distinct
 
         if (types.isEmpty) None else Some(types)
       }
--- a/macros/src/main/scala-3/enumeratum/EnumMacros.scala
+++ b/macros/src/main/scala-3/enumeratum/EnumMacros.scala
@@ -149,22 +149,14 @@ object EnumMacros:
 
           }
       }
-
       childTpr match {
         case Some(child) => {
+          val tpeSym = child.typeSymbol
           child.asType match {
-            case ct @ '[IsEntry[t]] => {
-              val tpeSym = child.typeSymbol
-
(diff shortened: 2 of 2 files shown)


Krever:
I found a serious regression when tried to use the snpashot in my codebase. I added handling a test case for it but please let me know if thats enough. I was a bit frightened when the existing tests didn't catch it.

lloydmeta (commented):
Interesting...

I feel like this likely means we should have some logic/rationale for what subclasses get picked up or not picked up based on their type param.

Krever:
Hey @lloydmeta, I just finished migrating my codebase, and this is the last blocker for me. Do you think it could be merged and released anytime soon?

lloydmeta:
Yes, I still think we should have good reason other than "it's complex" for what members get picked up based on the type param.

Krever:
Sure, but what would be the outline of the filtering logic? For me it seems that there is no inherent limitation to what members are being picked, as long as the final leas is a subtype. If you have some idea for a stricter filtering, I can try to implement it.

lloydmeta:
I'm thinking about the user perspective first; eg covariance? Self base trait? Etc. before worrying about implementation.

Krever:
So you're considering scenarios where the user defines an enum subtype inside the enum object but doesn't expect it to be picked up by the macro? 

From a variance perspective, I don't see a problem, I can imagine scenarios with either covariant, contravariant or invariant type parameter. 
Regarding the self-trait, I would probably need a more elaborate explanation.

But in the meantime, maybe from a different perspective: as this is a fix for regression between scala 2 and 3, maybe we should stick to the approach present there? 
I tried to look into scala 2 macros [here](https://github.com/lloydmeta/enumeratum/blob/6ac99af9cde646fa4651a4403fd372843ee3b663/macros/src/main/scala-2/enumeratum/EnumMacros.scala#L77) and the algorithm there seems to be very close (or the same) as one in this PR - traverse everything, collect subtypes. So maybe its ok?

lloydmeta:
Maintaining possibly odd behaviour is a non-goal between Scala 2 and Scala 3 versions. It's more constructive to make sure this actually makes sense going forward.


```scala
sealed trait Animal
sealed trait Dog extends Animal
sealed trait Cat extends Animal

sealed trait Vet[T] extends EnumEntry

object CatVets extends Enum[Vet[Cat]] {
  
  case object Joe extends Vet[Cat]
  case object CatHater extends Vet[Dog]

  lazy val values = findValues 
}


object DogVets extends Enum[Vet[Dog]] {
  
  case object John extends Vet[Dog]
  case object DogHater extends Vet[Cat]

  lazy val values = findValues 
}
```

That's the sort of thing I think we need to have a good answer for (well-reasoned + working implementation), before merging this.

Krever:
Oh yes, that's almost exactly the scenario I have in my codebase and it works well. When I was referring to subtype relationship, I meant relation to the type passed to `Enum[_]` hence
CatVets: 
1. `Joe <: Vet[Cat]` ✅ 
2. `CatHater <: Vet[Cat]` ❌

DogVets
1. `John <: Vet[Dog]` ✅ 
2. `DogHater <: Vet[Dog]` ❌

Would you like me to add test cases for those? And if so, is invariance enough or you think all 3 variances are better? (No preference on my side).

lloydmeta:
Concretely I think we should have a good reason for why `CatVets` in [my example](https://github.com/lloydmeta/enumeratum/pull/426#issuecomment-2904097318) should or should not work. IMO it should throw a compile time error.  Yes, tests would help to show this.

I don't actually have the time or energy to come up with a good set of reasons  (I've never had this type param requirement for enums), but I do have a strong preference that it should not (re)introduce surprising behaviour and if we can't think of a good reasoning, we should be as strict as possible.

Krever:
I share your preference and agree that the absence of a good reason is enough not to support a particular case. That said, I think this specific situation is tricky if we aim to prevent surprising behavior at compile time.

What’s easy — and already implemented — is simply not collecting invalid data.

What you’re suggesting (as I understand it) is to disallow defining objects with “invalid” supertypes altogether. That might be difficult or even impossible to enforce at compile time.

From the compiler’s perspective, the following two cases are essentially equivalent:

```
sealed trait Animal
sealed trait Dog extends Animal
sealed trait Cat extends Animal

sealed trait Vet[T] extends EnumEntry

object CatVets extends Enum[Vet[Cat]] {
  case object Joe extends Vet[Cat]
  case object CatHater extends Vet[Dog]

  lazy val values = findValues
}

object Cats extends Enum[Cat] {
  case object Garfield extends Cat
  case object Snoopy extends Dog

  lazy val values = findValues
}
```

Currently, there’s no protection against defining a `Dog` inside an `Enum[Cat]`.

What might be possible is emitting a warning (or error) when we see an object inside `Enum[A]` that:
* does not inherit from `A`
* but does share a common supertype with `A`

That would catch the “Dog among Cats” case. But:
* We’d need to handle common supertypes like `Any`, which could get messy.
* It wouldn’t help with mismatches involving type constructors.

An alternative idea would be to warn/error when an object inside `Enum[A]` doesn’t inherit from A, but shares the same **type constructor**:
* This avoids the `Any` issue
* It wouldn’t catch simple (non-generic) mistakes
* I’m also not confident I could implement this reliably — it’s quite outside my comfort zone

All that said, I share the general mindset of “be as restrictive as possible,” and I’m open to whatever checks you think are appropriate. But I’d really like to avoid a situation where perfect becomes the enemy of good. I just migrated a very large codebase over the past week, and I’d love to wrap that up. I also believe enumeratum is a great library — I’d much prefer to contribute here than fork it.

Let me know how you'd like to proceed. If you can define some semi-formal rule to implement, just let me know. Otherwise I'm not sure I will be able to do it myself.

lloydmeta:
I'd be fine with not having invalid subtypes (incl in type params), with tests showing that.

To set expectations though: while I appreciate your effort here  I maintain this lib in a small portion of my free time, most of which has been used up. If it does not line up with your needs and timeline, feel free to fork - it's open source for a reason.

How I would like to proceed remains unchanged - I'd like there to be good rationale as to why we think this should work this way, and have it documented and tested. If you work that out, with the above tests, great - otherwise I'm happy to let this PR sit until someone does.

lloydmeta:
Addressed by #459

PR #432, opened 2025-07-14
Doobie 1.0.0-RC10

sideeffffect:


sideeffffect:
Yes, that's true, sbt will complain, because it can't know that this particular update isn't breaking (it treats every upgrade in RC as breaking, because it can be). But that can be suppressed.

PR #437, opened 2025-10-30
Consider superclass constructor in ValueEnumMacros

### LINKED ISSUE #294: [ValueEnum] When using inheritance, Compile error: not all of the members have a literal/constant 'value:String' declaration
Version: `1.6.1`

```scala
import enumeratum.values.{ StringEnum, StringEnumEntry }

import scala.collection.immutable

sealed abstract class Day extends StringEnumEntry with Product with Serializable

object Day extends StringEnum[Day] {

  sealed abstract class Weekend(val value: String) extends Day
  sealed abstract class Workday(val value: String) extends Day

  def values: immutable.IndexedSeq[Day] = findValues

  final case object Sun extends Weekend("Sun")
  final case object Mon extends Workday("Mon")
  final case object Tue extends Workday("Tue")
  final case object Wed extends Workday("Wed")
  final case object Thu extends Workday("Thu")
  final case object Fri extends Workday("Fri")
  final case object Sat extends Weekend("Sat")

  final object Weekend extends StringEnum[Weekend] {

    def values: immutable.IndexedSeq[Weekend] = Day.values.collect { case day: Weekend => day }
  }

  final object Workday extends StringEnum[Workday] {

    def values: immutable.IndexedSeq[Workday] = Day.values.collect { case day: Workday => day }
  }
}

```

```sbt
[error] It looks like not all of the members have a literal/constant 'value:String' declaration, namely: object Sun, object Mon, object Tue, object Wed, object Thu, object Fri, object Sat.
[error] 
[error] This can happen if:
[error] 
[error] - The aforementioned members have their `value` supplied by a variable, or otherwise defined as a method
[error] 
[error] If none of the above apply to your case, it's likely you have discovered an issue with Enumeratum, so please file an issue :)
[error]          
[error]   val values: immutable.IndexedSeq[Day] = findValues
```

cryeo:
Resolve #294

The existing macro assumes that the `value` field is defined in one of the following ways:

- as a member declared inside each object (or case object); or
- as a constructor parameter on the `ValueEntryType` itself.

However, it fails in the following scenario (also described in #294):

- `ValueEntryType` is defined as a trait or abstract class without ctor params, and
- the actual `value` is supplied through an intermediate abstract class.

This PR addresses the issue by extending the macro logic to also consider superclass ctor, without breaking the guarantees about `value` literal-ness.

### DIFF
--- a/enumeratum-core/src/test/scala/enumeratum/values/ValueEnumSpec.scala
+++ b/enumeratum-core/src/test/scala/enumeratum/values/ValueEnumSpec.scala
@@ -194,6 +194,27 @@ class ValueEnumSpec extends AnyFunSpec with Matchers with ValueEnumHelpers {
         }
         """ should compile
       }
+
+      it("should compile for parameterless value enum entry with sealed abstract subclasses") {
+        """
+        sealed trait Day extends StringEnumEntry
+
+        object Day extends StringEnum[Day] {
+          sealed abstract class Weekend(val value: String) extends Day
+          sealed abstract class Workday(val value: String) extends Day
+
+          case object Sun extends Weekend("Sun")
+          case object Mon extends Workday("Mon")
+          case object Tue extends Workday("Tue")
+          case object Wed extends Workday("Wed")
+          case object Thu extends Workday("Thu")
+          case object Fri extends Workday("Fri")
+          case object Sat extends Weekend("Sat")
+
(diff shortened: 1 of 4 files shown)


lloydmeta (approved):
LGTM. Will merge when I have time to release, at the same time as #436 likely.

PR #466, opened 2026-03-14
Add zio-json support

agilesteel:


lloydmeta:
Hey thanks for sending this but I'd recommend you to maintain and publish this integration yourself as I don't use ZIO and am not looking to add more integration libs at the moment.

PR #467, opened 2026-03-25
Added singleton Put for doobie enums

felix-hedenstrom:
At $work we've have a lot of enums like 
```
sealed trait FooStatus extends EnumEntry.UpperSnakecase

object FooStatus extends enumeratum.Enum[FooStatus] with DoobieEnum[FooStatus]:
  val values = findValues

  case object Initiated        extends FooStatus
  case object Completed        extends FooStatus
```

which are used in fragments with `status = ${FooStatus.Initiated: FooStatus}`. We must manually widen the type in order for the `Write[FooStatus]` instance to be found.

This PR makes `Write`-instances available for all singleton values in an `enumeratum-doobie` Enum. We can now write
`status = ${FooStatus.Initiated}` which is a lot cleaner.

### DIFF
--- a/enumeratum-doobie/src/main/scala/enumeratum/DoobieEnum.scala
+++ b/enumeratum-doobie/src/main/scala/enumeratum/DoobieEnum.scala
@@ -27,6 +27,8 @@ import doobie.Meta
   */
 trait DoobieEnum[A <: EnumEntry] { this: Enum[A] =>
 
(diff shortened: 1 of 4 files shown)


lloydmeta (approved):
LGTM. Will merge when I have time to release as well.

lloydmeta:
Released in 1.9.7

